PR #29, opened 2026-05-27
Fix add-leetcode-items script env loading

### LINKED ISSUE #24: [Bug] Script `add-leetcode-items.ts` crashes with 'Cannot find module dotenv'
### Description
Running the `scripts/add-leetcode-items.ts` script fails with:
```
Error: Cannot find module 'dotenv'
Require stack:
- D:\leetcode\scripts\add-leetcode-items.ts
```

The script imports `dotenv` but it's not listed in `package.json` dependencies or devDependencies. Other scripts use the `--env-file=.env.local` flag with `tsx` which doesn't need `dotenv`.

### Steps to Reproduce
1. Run any command that triggers `scripts/add-leetcode-items.ts`
2. Script crashes immediately with MODULE_NOT_FOUND error

### Expected Behavior
Either:
1. Add `dotenv` as a dev dependency: `npm install -D dotenv`
2. Or refactor the script to use the `--env-file` pattern like other scripts in `package.json`

### Evidence
Error captured in `error.log` at the project root.

### Affected Files
- `scripts/add-leetcode-items.ts` — Import statement
- `package.json` — Missing dependency

---
**Note for Contributors:** Please create a branch named with this issue number and name to make it easier to identify (e.g., `git checkout -b <issue-number>-issue-name`).

Chris79OG:
## What
- Adds an `add-leetcode-items` npm script that runs `scripts/add-leetcode-items.ts` with the same `.env.local` loading pattern used by the other maintenance scripts.
- Adds `tsx` to devDependencies so those script runners are resolved from the project lockfile instead of transient `npx` installs.

## Why
Issue #24 reports a missing `dotenv` module while running `add-leetcode-items.ts`. Current `main` no longer imports `dotenv`, so the safer fix is to route this script through the existing `tsx --env-file=.env.local` pattern rather than adding a new env-loading dependency.

## Verification
- `npm ci --ignore-scripts`
- `node -e "const s=require('./package.json').scripts['add-leetcode-items']; if (s !== 'npx tsx --env-file=.env.local scripts/add-leetcode-items.ts') throw new Error(s); console.log(s)"`
- `npx tsc --noEmit --pretty false`
- `git diff --check`

Note: `npm run add-leetcode-items` now resolves the project script and fails early with `.env.local: not found` when no local env file exists, instead of a missing `dotenv` module. Full `npm run lint` is currently blocked by pre-existing repo-wide lint errors unrelated to this package metadata change.

Closes #24

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,6 +36,7 @@
         "eslint": "^9",
         "eslint-config-next": "16.1.6",
         "tailwindcss": "^4",
+        "tsx": "^4.22.3",
         "typescript": "^5"
       }
     },
@@ -83,7 +84,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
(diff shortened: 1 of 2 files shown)


Ixotic27 (approved):
✅ Looks great! Thank you for the contribution. Merging now.

PR #106, opened 2026-05-28
fix: enforce dynamic rendering and disable caching for presence API route

### LINKED ISSUE #104: 🐛 Bug: Coding presence API is statically cached causing sync failures
### Bug Description
There is a synchronization issue between the IDE extension and the website. When a user opens the website again after their initial setup, the website incorrectly shows that they are not coding, even when their IDE (like Antigravity) is open and active.

*(Note for contributors: You must be **signed in to the website** to properly test and understand this problem, and to verify if your fix actually resolves it!)*

### Root Cause
The `GET` endpoint at `src/app/api/presence/route.ts` is being statically cached by the Next.js App Router. Because the route handler does not read any dynamic properties (like `cookies()` or `headers()`), Next.js statically caches the response. This means the `Date.now()` calculation for the 5-minute heartbeat cutoff is completely stale, preventing the website from fetching live developer sessions on page load.

### Steps to Reproduce
1. Sign in to the website.
2. Open your IDE and ensure the extension is sending heartbeats.
3. Check the website to see the coding presence.
4. Close the website and reopen it later — the presence will be out of sync.

### Expected Behavior
The `/api/presence` route should dynamically evaluate `Date.now()` and fetch the latest active developers from the database every time it's called.

### Proposed Solution
Add `export const dynamic = 'force-dynamic';` to the top of `src/app/api/presence/route.ts` to ensure Next.js never caches this API route statically.

gc-MayankPun:
## What does this PR do?

Fixes a stale presence bug caused by Next.js statically caching the `GET /api/presence` route at build time. Because the handler had no dynamic inputs (no `cookies()`, `headers()`, etc.), Next.js froze the `Date.now()` call used for the 5-minute heartbeat cutoff, meaning the response never reflected live session data after the initial build.

Two changes fix this:
- Added `export const dynamic = "force-dynamic"` to force per-request evaluation of the route.
- Replaced the `Cache-Control` header value `s-maxage=10, stale-while-revalidate=20` with `no-store` to prevent CDNs and browsers from serving stale presence data.

## Related issue

Fixes #104

## Screenshots

N/A no visual changes. Verified by signing in, triggering IDE heartbeats via VS Code, and confirming the website correctly reflects coding presence on page reload.

## Checklist
- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/app/api/presence/route.ts
+++ b/src/app/api/presence/route.ts
@@ -1,6 +1,8 @@
 import { NextResponse } from "next/server";
 import { getSupabaseAdmin } from "@/lib/supabase";
 
+export const dynamic = "force-dynamic";
+
 export async function GET() {
   const sb = getSupabaseAdmin();
   const cutoff = new Date(Date.now() - 5 * 60_000).toISOString();
@@ -47,7 +49,7 @@ export async function GET() {
     { count: developers.length, developers },
     {
       headers: {
-        "Cache-Control": "s-maxage=10, stale-while-revalidate=20",
+        "Cache-Control": "no-store",
       },
     },
   );


gc-MayankPun:
@Ixotic27 , could you please add the `gssoc:approved` label instead of `gssoc:verified` / `GSSOC 26`?
The leaderboard tracker only counts PRs with the approved label, otherwise the points don’t get reflected 😅
Thanks!

PR #137, opened 2026-05-30
fix(checkin): make streak reward grant atomic

### LINKED ISSUE #114: [Bug]: Non-atomic streak reward grant in `/api/checkin` allows duplicate item delivery under concurrent requests
### Bug Description

`grantStreakReward` in `src/app/api/checkin/route.ts` implements a check-then-insert pattern across three separate, non-transactional database operations:

1. Read `streak_rewards` to check if milestone was already claimed (`maybeSingle()`)
2. Insert into `purchases`
3. Insert into `streak_rewards`

There is no database transaction, no `upsert` with conflict handling on `streak_rewards`, and no application-level lock. Two concurrent POST requests to `/api/checkin` — which can occur due to a client retry, a browser page reload mid-flight, or React StrictMode double-invoke in development — can both pass the `maybeSingle()` existence check before either has committed the `streak_rewards` row.

Both requests then proceed to insert into `purchases` with the same `provider_tx_id` (`streak_reward_{milestone}_{developerId}`). If the purchases table does not have a unique constraint on `provider_tx_id`, the user receives the item twice. If it does, one insert throws and the error is silently swallowed while the milestone record is never written — leaving the user in a state where they have no `streak_rewards` record for that milestone and will trigger the flow again on their next checkin.

The 1-req/5s rate limit reduces, but does not close, this window — parallel in-flight requests are not serialized by it.

### Acknowledgements

Srejoye:
## What does this PR do?

Fixes a race condition in `grantStreakReward` inside `src/app/api/checkin/route.ts` where two concurrent POST requests to
`/api/checkin` could both pass the `maybeSingle()` existence check before either had committed a `streak_rewards` row — resulting in a user receiving a duplicate item or the milestone record never being written.

**Changes:**
- `streak_rewards` insert replaced with `upsert` on `onConflict: "developer_id,milestone"` with `ignoreDuplicates: true`
- Added a guard: only proceed to grant the purchase if the upsert actually inserted a new row
- `purchases` insert replaced with `upsert` on `provider_tx_id` as a secondary safety net

Only 1 file changed: `src/app/api/checkin/route.ts`

## Related issue

Fixes #114

## Checklist

### DIFF
--- a/src/app/api/checkin/route.ts
+++ b/src/app/api/checkin/route.ts
@@ -50,8 +50,22 @@ async function grantStreakReward(
       ? unowned[Math.floor(Math.random() * unowned.length)]
       : tier.pool[Math.floor(Math.random() * tier.pool.length)]; // fallback: grant anyway
 
+    // Record the reward first — upsert guards against concurrent duplicate grants.
+    // If another request already inserted this milestone, do nothing and skip.
+    const { data: rewardInserted } = await sb
+      .from("streak_rewards")
+      .upsert(
+        { developer_id: developerId, milestone: tier.milestone, item_id: itemId },
+        { onConflict: "developer_id,milestone", ignoreDuplicates: true }
+      )
+      .select("id")
+      .maybeSingle();
+
+    // Only grant the item if we actually inserted the reward row (not a duplicate)
+    if (!rewardInserted) continue;
+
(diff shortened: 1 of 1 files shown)


Ixotic27 (approved):

PR #139, opened 2026-05-30
fix(queue): guard flush() against concurrent invocations

### LINKED ISSUE #116: [Bug]: Concurrent `flush()` invocations in VS Code extension heartbeat queue can cause heartbeat loss on send failure
### Bug Description

`flush()` in `packages/vscode-extension/src/api/queue.ts` is an `async` function with no in-flight guard. Two independent callers can invoke it concurrently:

- `setInterval(flush, FLUSH_INTERVAL_MS)` (every 30 seconds)
- `enqueue()`, which calls `flush()` immediately when the queue transitions from empty to non-empty (line 28)

Because JavaScript is single-threaded, the `splice` on line 51 cannot race. However, the destructive pattern occurs *across the await boundary*: Flush A splices a batch and starts `await sendHeartbeats(batch)`. While awaiting, new heartbeats arrive. The interval fires, Flush B starts, splices those new heartbeats, sends them, and calls `persist()` — writing a queue state that does not include Flush A's batch. Flush A's HTTP call then fails; it calls `queue.unshift(...batch)`, which correctly restores the in-memory queue, but the persisted state in `globalState` (written by Flush B) does not include them. If VS Code is closed before the next flush, those heartbeats are permanently lost from storage even though they were never successfully sent.

The inverse also occurs: if the interval fires while `enqueue`'s immediate flush is in-flight and new items arrive between the splice and the await resolution, both flushes attempt to drain what they think is the full queue, and the interval flush may send items that Flush A is about to retry — resulting in duplicates at the server.

### Acknowledgements

hariom888:
## What does this PR do?

Fixes a heartbeat loss bug in the VS Code extension caused by concurrent `flush()` invocations.

`flush()` was `async` with no in-flight guard. Two independent callers could invoke it simultaneously:
- `setInterval(flush, FLUSH_INTERVAL_MS)` every 30 seconds
- `enqueue()` calling `flush()` immediately when the queue goes from empty to non-empty

Across the `await` boundary: Flush A splices a batch and starts `await sendHeartbeats()`. While awaiting, Flush B runs, splices new items, succeeds, and calls `persist()` — writing a queue state without Flush A's batch. Flush A then fails, unshifts its batch back into memory, but the persisted state no longer includes it. If VS Code closes before the next interval tick, those heartbeats are permanently lost.

**Fix:**
- Added `let isFlushing = false` guard
- `flush()` returns early if already in-flight
- `persist()` moved into `finally` block so storage always reflects   correct in-memory state after the await settles
- `enqueue()` skips immediate flush if one is already in-flight

1 file changed: `packages/vscode-extension/src/api/queue.ts`

## Related issue

Fixes #116

## Checklist

### DIFF
--- a/packages/vscode-extension/src/api/queue.ts
+++ b/packages/vscode-extension/src/api/queue.ts
@@ -3,9 +3,10 @@ import { sendHeartbeats } from "./client";
 import { FLUSH_INTERVAL_MS, MAX_BATCH_SIZE, QUEUE_STORAGE_KEY } from "../constants";
 import type { RawHeartbeat } from "../privacy/sanitizer";
 
-const queue: RawHeartbeat[] = [];
+let queue: RawHeartbeat[] = [];
 let flushTimer: ReturnType<typeof setInterval> | undefined;
 let globalState: vscode.Memento;
+let isFlushing = false;
 
 export function initQueue(context: vscode.ExtensionContext) {
   globalState = context.globalState;
@@ -24,7 +25,8 @@ export function enqueue(heartbeat: RawHeartbeat) {
   queue.push(heartbeat);
   persist();
   // Flush immediately when first heartbeat arrives (instant feedback)
-  if (wasEmpty) {
+  // Skip if a flush is already in-flight — interval will pick it up
+  if (wasEmpty && !isFlushing) {
     flush();
   }
 }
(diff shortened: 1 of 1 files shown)


Ixotic27 (approved):

PR #174, opened 2026-05-31
fix(shader): prevent stale building uniforms

### LINKED ISSUE #156: [Bug]: ShaderMaterial created with empty dependency array causes stale uniforms
## Description
In `src/components/InstancedBuildings.tsx` (line 224), the shader material is memoized with an empty dependency array:

```typescript
const material = useMemo(() => {
  return new THREE.ShaderMaterial({ /* ... */ });
}, []); // ← empty deps!
```

This means the material is created **once** and never recreated. If the atlas texture or colors change (e.g., theme switch), the material's initial uniform values become stale. While there's a `useEffect` on line 245 that manually updates some uniforms, it's error-prone and doesn't cover all uniforms.

### Fix
Either:
1. Add proper dependencies to the `useMemo`, or
2. Document why the empty deps are intentional and ensure ALL uniforms are handled in the follow-up `useEffect`

### Checklist
- [ ] Audit all shader uniforms for staleness
- [ ] Fix dependency array or add comprehensive manual updates
- [ ] Test theme switching with the fix
- [ ] Provide **screenshots** of correct rendering after theme switch

Shreeya1207:
## What does this PR do?

This PR fixes a potential stale uniform issue in `src/components/InstancedBuildings.tsx`.

Previously, the building `ShaderMaterial` was memoized with an empty dependency array:

```tsx
const material = useMemo(() => {
  return new THREE.ShaderMaterial({...});
}, []);
```

Because the material was only created once, uniforms initialized from dynamic values such as:

* `atlasTexture`
* `colors.roof`
* `colors.face`
* `dimOpacity`
* `dimEmissive`
* `cityEnergy`

could become stale when those values changed (for example during theme changes or visual state updates).

### Changes Made

* Added all dynamic shader inputs to the `useMemo` dependency array.
* Removed the manual uniform synchronization effect.
* Ensured the material is recreated whenever its initialization values change.
* Prevented outdated texture and color values from being retained by the shader.

This keeps the shader uniforms in sync with the latest theme and rendering state while simplifying the update logic.

## Related issue

Fixes #156

## Checklist

* [x] Audited shader uniforms for potential staleness
* [x] Added proper dependencies to the `useMemo` hook
* [x] Removed redundant manual uniform synchronization logic
* [x] Verified the application loads successfully after the change
* [x] Ensured no new runtime errors were introduced
* [ ] Added screenshots (to be attached if required by maintainers)
* [ ] Tested theme switching visually (environment limitation)

### DIFF
--- a/src/components/InstancedBuildings.tsx
+++ b/src/components/InstancedBuildings.tsx
@@ -220,7 +220,6 @@ export default memo(function InstancedBuildings({
   }, [buildings]);
 
   const geo = useMemo(() => new THREE.BoxGeometry(1, 1, 1), []);
-
   const material = useMemo(() => {
     return new THREE.ShaderMaterial({
       uniforms: {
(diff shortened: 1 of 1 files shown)


Ixotic27 (approved):

Thank you for your contribution @Shreeya1207! 🚀

PR #179, opened 2026-05-31
fix: atomic optimistic lock on redeem code used_count — prevent TOCTOU over-redemption 

### LINKED ISSUE #162: [Bug]: TOCTOU Race Condition on Redeem Code `used_count` — Over-Redemption Possible
### Description

The redeem code flow performs a non-atomic check-then-act sequence:

1. `SELECT` the code row and read `used_count` and `max_uses` into application memory.
2. Validate `used_count < max_uses` in JavaScript.
3. INSERT a purchase record for the user.
4. UPDATE (or DELETE) `redeem_codes` with `used_count + 1`.

Steps 1 and 4 are not wrapped in a database transaction, and step 4 uses an absolute value (`used_count + 1` computed in JS from the snapshot taken in step 1) rather than a server-side atomic increment (`used_count = used_count + 1`). Under concurrent redemptions — two requests that both pass the validation check before either has committed their update — the counter ends up written twice with the same value, effectively consuming only one slot despite two items being granted.

For single-use codes (`max_uses = 1`) this allows multiple users to simultaneously redeem the code before the row is deleted, granting the item to every concurrent requester.

### Affected Files

- `src/app/api/shop/redeem/route.ts` — entire redeem flow

### Steps to Reproduce

1. Issue a redeem code with `max_uses = 1`.
2. Fire two simultaneous `POST /api/shop/redeem` requests with that code from two different authenticated sessions.
3. Observe both requests succeed and both users receive the item.

### Expected Behavior

Limited-edition or single-use promotional items can be duplicated across any number of concurrent claimants, undermining scarcity guarantees.

_No response_

### Browser

_No response_

### Acknowledgements

Srejoye:
## What does this PR do?

Fixes a TOCTOU race condition in `POST /api/shop/redeem` where two concurrent requests could both pass the `used_count < max_uses` JavaScript check before either committed their update, causing both to receive the item while `used_count` only incremented once.

The non-atomic sequence was:
1. `SELECT` → read `used_count` snapshot
2. JS validates `used_count < max_uses`
3. `INSERT` purchase
4. `UPDATE used_count = snapshot + 1` — blind overwrite, race lost here

**Fix:**

Replaced the blind `UPDATE` with an optimistic-lock conditional update:

```typescript
.update({ used_count: redeemCode.used_count + 1 })
.eq("id", redeemCode.id)
.eq("used_count", redeemCode.used_count) // guard — concurrent request gets 0 rows
```

If the `used_count` has already been incremented by a concurrent request, the `WHERE used_count = <snapshot>` condition matches 0 rows and `maybeSingle()` returns `null`. The handler immediately returns 409, and no purchase is inserted for the losing request.

Also changed `provider_tx_id` from `code` to `${code}:${dev.id}` so the unique constraint on `(developer_id, provider_tx_id)` serves as a last-resort idempotency guard at the DB layer.

**Files changed:**
- `src/app/api/shop/redeem/route.ts` — optimistic lock on `used_count` update, idempotent `provider_tx_id`

## Related issue

Fixes #162

## Checklist


Ixotic27 (approved):

PR #180, opened 2026-05-31
fix: atomic optimistic lock on points deduction — prevent double-spend in buy-with-points

### LINKED ISSUE #163: [Bug]: Non-Atomic Points Deduction in `buy-with-points` Allows Double-Spend
### Description

The points purchase flow reads the developer's current `points` balance into application memory, computes the new balance in JavaScript, and issues a blind `UPDATE developers SET points = <computed_value>`. There is no database-level concurrency guard (no `WHERE points >= price` condition, no row-level lock, no transaction spanning the read and write).

If two concurrent requests both read the same balance before either write commits, both will compute `balance - price` from the same snapshot and both writes will succeed — with the second overwriting the first. The net result is the user spends `price` points once but receives two items (or spends zero points if the two purchases happen to write the same value back).

The comment `// We'll use a transactionally-safe approach` in the code is incorrect; the two statements (UPDATE then INSERT) are independent round-trips with no transaction or optimistic-concurrency guard.

Additionally, the "rollback" path on purchase insert failure restores `dev.points` (the snapshot value) rather than adding `price_points` back to the current DB value, which can corrupt the balance if any other concurrent write landed between the deduct and the rollback.

### Affected Files

- `src/app/api/shop/buy-with-points/route.ts`

### Steps to Reproduce

1. **Preferred — DB-side RPC:** Add a Postgres function `buy_item_with_points(p_developer_id, p_item_id)` that: `SELECT ... FOR UPDATE` the developer row, validates balance, inserts the purchase, and decrements points atomically — all inside one transaction. The API handler calls this and maps the result.

2. **Simpler — Conditional UPDATE:** Change the deduction to:
```sql
   UPDATE developers SET points = points - :price WHERE id = :dev AND points >= :price
```
   Check `rowsAffected`; if 0, the balance was insufficient (race lost). Make the purchase insert idempotent (`ON CONFLICT ... DO NOTHING` on `(developer_id, item_id)` for non-consumables). Drop the manual rollback entirely.

The broken rollback path (`SET points = dev.points`) must be replaced with `SET points = points + price` regardless of which approach is chosen.

- Add a concurrent double-purchase test: fire two simultaneous buy-with-points requests for the same item from the same user; assert exactly one succeeds and points decrease by exactly `price` once.
- Test that a purchase insert failure leaves the balance unchanged (not corrupted).
- Existing single-user purchase and "already owned" tests should be unaffected.

### Expected Behavior

A user can trigger two simultaneous purchase requests for the same or different items and receive both while only being charged for one, or potentially receive items for free if the rollback overwrites a legitimate balance update.

_No response_

### Browser

_No response_

### Acknowledgements

Srejoye:
## What does this PR do?

Fixes a non-atomic race condition in `POST /api/shop/buy-with-points` where two concurrent requests could both read the same `points` balance snapshot, both pass the JS `>= price` check, and both successfully write `balance - price` back — the second overwriting the first, granting two items for the cost of one.

The broken sequence was:
1. `SELECT points` → JS snapshot
2. JS validates `snapshot >= price`
3. `UPDATE points = snapshot - price` — blind overwrite, no guard
4. `INSERT` purchase

Two concurrent requests reading the same snapshot both write the same value back; the counter effectively decrements once despite two grants.

A secondary bug: the rollback on purchase insert failure did `SET points = dev.points` (restoring the original snapshot), which would **overwrite** any legitimate concurrent write that landed between the deduct and the rollback, corrupting the balance.

**Fix:**

Replaced the blind `UPDATE` with an optimistic-lock conditional update:

```typescript
.update({ points: dev.points - item.price_points })
.eq("id", dev.id)
.gte("points", item.price_points)  // balance still sufficient
.eq("points", dev.points)          // snapshot must still match — race loses here
.select("points")
.maybeSingle()
```

If the snapshot no longer matches (concurrent write landed first), `maybeSingle()` returns `null` and the handler returns 409. The winning request is the only one that lands the deduction.

Also fixed the rollback path: instead of restoring the snapshot value, it now adds `price_points` back to the current committed DB value (`deducted.points + item.price_points`), making it safe regardless of any concurrent writes.

**Files changed:**
- `src/app/api/shop/buy-with-points/route.ts` — optimistic lock on points deduction, safe additive rollback, correct `points_remaining` in response

## Related issue

Fixes #163

## Checklist

### DIFF
--- a/src/app/api/shop/buy-with-points/route.ts
+++ b/src/app/api/shop/buy-with-points/route.ts
@@ -78,15 +78,20 @@ export async function POST(request: Request) {
         return NextResponse.json({ error: "Not enough points" }, { status: 403 });
     }
 
-    // 4. Atomic transaction via RPC or sequence
-    // We'll use a transactionally-safe approach: deduct points then insert purchase
-    const { error: deductError } = await admin
+    // 4. Atomic conditional deduction — only succeeds if balance is still sufficient
+    // Optimistic lock: WHERE points >= price_points ensures concurrent requests
+    // that both passed the JS balance check cannot both land this update
+    const { data: deducted, error: deductError } = await admin
         .from("developers")
         .update({ points: dev.points - item.price_points })
-        .eq("id", dev.id);
+        .eq("id", dev.id)
+        .gte("points", item.price_points) // guard — race condition loses here
+        .eq("points", dev.points)         // optimistic lock on exact snapshot value
+        .select("points")
+        .maybeSingle();
 
-    if (deductError) {
-        return NextResponse.json({ error: "Failed to deduct points" }, { status: 500 });
+    if (deductError || !deducted) {
+        return NextResponse.json({ error: "Not enough points or a concurrent purchase already deducted your balance. Please try again." }, { status: 409 });
     }
 
     const { data: purchase, error: purchaseError } = await admin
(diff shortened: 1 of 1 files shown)


Ixotic27 (approved):

PR #182, opened 2026-05-31
fix: fail closed on raid limit/peace shield checks — remove silent exception swallow

### LINKED ISSUE #164: [Bug]: Raid Rate-Limit and Peace Shield Checks Silently Swallowed — Limits Become Unenforceable
### Bug Description

The entire block that enforces the three raid constraints — daily raid cap (`MAX_RAIDS_PER_DAY = 3`), the 2-hour peace shield on the defender, and the weekly per-pair cooldown — is wrapped in a single `try/catch` that logs a warning and continues execution on any database error:

```typescript
try {
  // daily cap check
  // peace shield check
  // weekly cooldown check
} catch (err) {
  console.warn("[app/api/raid/execute/route.ts] non-critical error:", err);
}
// execution continues regardless
```

If the `raids` table is temporarily unavailable, a query times out, or a Supabase connection limit is hit, all three enforcement checks are silently skipped and the raid proceeds. A user can exploit this by triggering the condition (e.g., via connection pool exhaustion during a burst) or by deploying to an environment where the `raids` table hasn't been migrated yet — a scenario the comment explicitly anticipates ("raids table may not exist yet").

The fallback path below the block falls straight through to the `execute_raid` RPC call, which does insert the raid row, meaning the limits are bypassed while the raid outcome is still recorded and XP/tags are granted.

**Affected Modules / Files**

- `src/app/api/raid/execute/route.ts` — enforcement try/catch block
- `supabase/migrations/015_raid_system.sql` — table definition reference

### Steps to Reproduce

**Proposed Fix Approach**

1. **Preferred — Move checks into the RPC:** The `execute_raid` Postgres function should atomically enforce all three limits inside the transaction. Violations return a typed error code rather than throwing. This eliminates the application-layer race and the need for the try/catch entirely.

2. **Minimal fix — Fail closed:** Remove the try/catch wrapper. Let DB errors surface as 500s to the client rather than silently granting the raid. Each check that depends on schema existence can be individually guarded with a table-existence check or a separate migration-state flag.

3. **If the missing-table tolerance must be preserved:** Check specifically for the "table does not exist" error code (`42P01`) and only continue in that case; re-throw all other errors.

- Add a test that mocks the `raids` table query to throw a connection error and asserts the endpoint returns a 5xx rather than proceeding.
- Add tests for each of the three limits (daily cap, peace shield, weekly cooldown) asserting they return 429 independently.
- Existing happy-path raid tests should be unaffected.

### Expected Behavior

- A single user can execute unlimited raids per day against any target, bypassing the 3/day cap.
- A defender's 2-hour peace shield can be ignored, allowing repeated raids on the same target.
- The weekly per-pair cooldown can be bypassed.

_No response_

### Browser

_No response_

### Acknowledgements

hariom888:
## What does this PR do?

Fixes a critical security bug where all three raid enforcement checks — daily cap, 2-hour peace shield, and weekly per-pair cooldown — were wrapped in a single `try/catch` that silently swallowed any database error and allowed the raid to proceed unconditionally.

The original code:
```typescript
try {
  // daily cap check
  // peace shield check
  // weekly cooldown check
} catch (err) {
  console.warn("non-critical error:", err);
}
// raid proceeds regardless
```

Any DB error (timeout, connection pool exhaustion, missing table) bypassed all three limits simultaneously while still granting XP and inserting the raid row via `execute_raid`.

**Fix:**

Replaced the single swallowing try/catch with individual inline error checks on each query. Each check now **fails closed** — returning HTTP 500 on any unexpected DB error rather than silently proceeding.

The only tolerated error is Postgres code `42P01` (table does not exist), which preserves the original intent of allowing the route to work before the `raids` table migration has run. All other errors surface as 500s.

The peace shield check is pure in-memory (no DB call) and needed no change.

**Files changed:**
- `src/app/api/raid/execute/route.ts` — replaced try/catch with inline per-check error handling, fail-closed on unexpected DB errors

## Related issue

Fixes #164

## Checklist

### DIFF
--- a/src/app/api/raid/execute/route.ts
+++ b/src/app/api/raid/execute/route.ts
@@ -105,42 +105,50 @@ export async function POST(request: Request) {
   const todayStart = new Date();
   todayStart.setHours(0, 0, 0, 0);
 
-  try {
-    const { count: raidsToday } = await admin
-      .from("raids")
-      .select("id", { count: "exact", head: true })
-      .eq("attacker_id", attacker.id)
-      .gte("created_at", todayStart.toISOString());
+  // Daily cap check
+  const { count: raidsToday, error: dailyCheckError } = await admin
+    .from("raids")
+    .select("id", { count: "exact", head: true })
+    .eq("attacker_id", attacker.id)
+    .gte("created_at", todayStart.toISOString());
 
-    if ((raidsToday ?? 0) >= MAX_RAIDS_PER_DAY) {
-      return NextResponse.json({ error: "Daily raid limit reached" }, { status: 429 });
+  if (dailyCheckError) {
+    // Only tolerate missing-table errors (schema not yet migrated); fail closed on all others
+    if (!dailyCheckError.message?.includes("42P01")) {
+      console.error("[raid/execute] daily cap check failed:", dailyCheckError.message);
+      return NextResponse.json({ error: "Raid temporarily unavailable" }, { status: 500 });
     }
-
-    // Check 2-hour peace shield on defender
-    if (defender.last_raided_at) {
-      const shieldExpires = new Date(new Date(defender.last_raided_at).getTime() + 2 * 60 * 60 * 1000);
-      if (new Date() < shieldExpires) {
-        return NextResponse.json({ error: "Target has an active Peace Shield" }, { status: 429 });
-      }
-    }
-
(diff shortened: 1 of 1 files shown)


Ixotic27 (approved):

PR #183, opened 2026-05-31
fix: pass isMobile flag to trackDailyMission — mobile users can now complete their assigned missions

### LINKED ISSUE #165: [Bug]: Server-Side `trackDailyMission` Always Uses Desktop Mission Set — Mobile Users Cannot Complete `desktopOnly` Missions But Also Cannot Complete Their Actual Assigned Missions
### Bug Description

`getDailyMissions` is a deterministic function that accepts an `isMobile` flag. When `isMobile = true`, it filters out missions marked `desktopOnly` (currently `fly_score_50` and `fly_score_150`) before running the Fisher-Yates shuffle, producing a different 3-mission set than the desktop variant for the same `developerId + date`.

The server-side `trackDailyMission` helper (called from `checkin`, `kudos`, `visit`, and `raid/execute` routes) always calls `getDailyMissions(developerId, today)` **without the `isMobile` parameter**, hardcoding `isMobile = false`. This means it always computes the desktop mission set.

For a mobile user, whose client was assigned a mobile-specific mission set, a server-side event (e.g., visiting a building) will call `getDailyMissions(id, today, false)`. If the desktop shuffle assigned `fly_score_50` in a slot that the mobile shuffle filled with `visit_building`, then `trackDailyMission(id, "visit_building")` returns early (`mission not assigned today, skip`) and progress is never recorded, even though the user can see the mission on their screen and is performing the required action.

The `/api/dailies/claim` route correctly accepts `isMobile` from the request body, but the upstream tracking calls that feed the `daily_mission_progress` table do not, creating a desynchronization between what is displayed and what is tracked.

### Affected Files

- `src/lib/dailies.ts` — `trackDailyMission` function (line 103)
- `src/app/api/interactions/visit/route.ts` — calls `trackDailyMission` without mobile flag
- `src/app/api/interactions/kudos/route.ts` — same
- `src/app/api/checkin/route.ts` — same
- `src/app/api/raid/execute/route.ts` — same

### Steps to Reproduce

Two options:

1. **Store `isMobile` on the session/developer row:** When the client first fetches dailies (`GET /api/dailies`), persist `isMobile` (from the `User-Agent` or explicit client signal) to the `developers` table or a session store. `trackDailyMission` then reads this flag from the DB instead of accepting it as a parameter. Adds one DB column and one read per tracking call, but keeps the API surface clean.

2. **Pass `isMobile` in each tracking request:** Update `trackDailyMission` to accept and forward `isMobile`. Routes that originate from client requests extract it from the `User-Agent` header (or a request body field) and forward it. This is lower DB overhead but requires touching every call site.

Either way, the `trackDailyMission` call in `src/app/api/raid/execute/route.ts` for `attempt_battle` / `win_battle` is a server-initiated event with no client context; a stored flag (option 1) is required there.

- Add a unit test asserting that `getDailyMissions(id, date, false)` and `getDailyMissions(id, date, true)` can differ for at least one seed, and that `trackDailyMission` with the wrong `isMobile` value returns early when it should not.
- Add an integration test simulating a mobile session completing `visit_building` and asserting `daily_mission_progress` is updated.
- Existing desktop daily-completion tests should be unaffected.

### Expected Behavior

Mobile users assigned a non-`desktopOnly` mission that lands in a shuffle slot that differs between desktop and mobile sets will have their progress silently ignored. They see the mission as active on-screen but can never complete it, permanently blocking daily reward collection for that day.

_No response_

### Browser

_No response_

### Acknowledgements

hariom888:
## What does this PR do?

Fixes a bug where mobile users could see their daily missions on screen but could never complete them because server-side progress tracking always computed the **desktop** mission set.

`getDailyMissions` is deterministic but platform-sensitive: when `isMobile = true`, the two `desktopOnly` missions (`fly_score_50`,
`fly_score_150`) are excluded before the Fisher-Yates shuffle, producing a different 3-mission set for the same `developerId + date`. Server-side `trackDailyMission` was hardcoded to `isMobile = false`, so it looked up the desktop set. If the desktop shuffle placed a `desktopOnly` mission in a slot that the mobile shuffle filled with e.g. `visit_building`, then `trackDailyMission(id, "visit_building")` returned early with `"not assigned today, skip"` — silently ignoring the user's action.

**Fix:**

Added `isMobile?: boolean` to the `trackDailyMission` `extra` parameter and forwarded it into `getDailyMissions`. Each client-facing route (`checkin`, `visit`, `kudos`) detects the platform from the `User-Agent` header and passes it through. The `raid/execute` route uses `false` explicitly since raid missions are never `desktopOnly`.

**Files changed:**
- `src/lib/dailies.ts` — `trackDailyMission` accepts `extra.isMobile`, forwards to `getDailyMissions`
- `src/app/api/checkin/route.ts` — detect mobile from User-Agent, pass to `trackDailyMission`
- `src/app/api/interactions/visit/route.ts` — same
- `src/app/api/interactions/kudos/route.ts` — same
- `src/app/api/raid/execute/route.ts` — explicit `isMobile: false` (raid missions are never desktopOnly, no User-Agent context available)

## Related issue

Fixes #165

## Checklist

### DIFF
--- a/src/app/api/checkin/route.ts
+++ b/src/app/api/checkin/route.ts
@@ -90,7 +90,7 @@ async function fetchWeeklyContributions(login: string): Promise<number | null> {
   return fetchLeetCodeWeeklySubmissions(login);
 }
 
-export async function POST() {
+export async function POST(request: Request) {
   const supabase = await createServerSupabase();
   const {
     data: { user },
@@ -146,7 +146,8 @@ export async function POST() {
 
   // Track activity
   await touchLastActive(dev.id);
-  await trackDailyMission(dev.id, "checkin");
+  const isMobile = /Mobi|Android|iPhone|iPad/i.test(request.headers.get("user-agent") ?? "");
+  await trackDailyMission(dev.id, "checkin", { isMobile });
 
   // Detect streak broken: previous streak was >= 7, now reset to 1, and freeze didn't save them
   const previousStreak = dev.app_streak ?? 0;
--- a/src/app/api/checkin/route.ts
+++ b/src/app/api/checkin/route.ts
@@ -146,8 +146,7 @@ export async function POST(request: Request) {
 
   // Track activity
   await touchLastActive(dev.id);
-  const isMobile = /Mobi|Android|iPhone|iPad/i.test(request.headers.get("user-agent") ?? "");
-  await trackDailyMission(dev.id, "checkin", { isMobile });
+  await trackDailyMission(dev.id, "checkin");
 
   // Detect streak broken: previous streak was >= 7, now reset to 1, and freeze didn't save them
   const previousStreak = dev.app_streak ?? 0;
--- a/src/app/api/dailies/route.ts
+++ b/src/app/api/dailies/route.ts
@@ -56,11 +56,11 @@ export async function GET(request: Request) {
       title: m.title,
       description: m.description,
       threshold: m.threshold,
+      desktopOnly: m.desktopOnly ?? false,
       progress: prog?.progress ?? 0,
       completed: prog?.completed ?? false,
     };
   });
-
   const completedCount = missionData.filter((m) => m.completed).length;
   const allCompleted = completedCount === 3;
   const alreadyClaimedToday = dev.last_dailies_date === today;
--- a/src/app/api/interactions/kudos/route.ts
+++ b/src/app/api/interactions/kudos/route.ts
@@ -89,8 +89,9 @@ export async function POST(request: Request) {
 
   // Track activity
   await touchLastActive(giver.id);
-  await trackDailyMission(giver.id, "give_kudos");
-  await trackDailyMission(giver.id, "give_kudos_3");
+  const isMobile = /Mobi|Android|iPhone|iPad/i.test(request.headers.get("user-agent") ?? "");
+  await trackDailyMission(giver.id, "give_kudos", { isMobile });
+  await trackDailyMission(giver.id, "give_kudos_3", { isMobile });
 
   // Only increment + feed if the insert actually happened (no conflict)
   if (!insertError) {
--- a/src/app/api/interactions/kudos/route.ts
+++ b/src/app/api/interactions/kudos/route.ts
@@ -89,9 +89,8 @@ export async function POST(request: Request) {
 
   // Track activity
   await touchLastActive(giver.id);
-  const isMobile = /Mobi|Android|iPhone|iPad/i.test(request.headers.get("user-agent") ?? "");
-  await trackDailyMission(giver.id, "give_kudos", { isMobile });
-  await trackDailyMission(giver.id, "give_kudos_3", { isMobile });
+  await trackDailyMission(giver.id, "give_kudos");
+  await trackDailyMission(giver.id, "give_kudos_3");
 
   // Only increment + feed if the insert actually happened (no conflict)
   if (!insertError) {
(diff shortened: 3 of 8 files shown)


Ixotic27 (changes requested):
Hi @hariom888! Thank you for this PR and for identifying why mobile tracking was failing!

However, there is a critical logic flaw with generating a different 3-mission set based on the `User-Agent` header. Because LeetCode City uses a stateless, deterministic PRNG to assign daily missions on the fly (to save database space), removing items from the pool *before* the shuffle changes the array size, which completely scrambles the math of the shuffle.

This means if a player logs in on Desktop, gets 3 missions, and later switches to Mobile in the same day, their `User-Agent` changes, the pool shrinks, the PRNG shifts, and their assigned missions for the day will completely re-roll! They will lose access to the missions they were working on.

**How to fix:**
Because the daily missions are stateless, the array logic **must remain perfectly identical** on all devices. Instead of filtering the pool in `getDailyMissions`, we should:

1. Always assign the same missions to everyone (do not pass or use `isMobile` in `getDailyMissions` or the tracking routes).
2. When the user is on their phone and looks at their daily missions, if they are assigned a `desktopOnly` mission (like the flying task), we should simply show a UI message/badge on that mission saying: *"This mission can only be completed on a PC, not on a phone."*

Could you please update this PR to revert the backend `User-Agent` logic, and instead implement this UI message in the frontend React component? Thank you!

Ixotic27:
Also, when you push the update, could you please include some screenshots (or preferably a short video) of how the message looks when accessing the task on a mobile view? We'd love to check out the styling, layout, and how the message works in action before merging. Thank you!

hariom888:
Hi @Ixotic27! Thank you for the detailed feedback. I've updated the PR with the following changes:

**Backend (reverted):**
- Removed `isMobile` from `trackDailyMission` signature and all call sites (`checkin`, `visit`, `kudos`, `raid/execute`)
- `getDailyMissions` now always uses the full pool on the server — mission sets are identical across all devices, preserving PRNG determinism

**Frontend (new approach):**
- Added `desktopOnly?: boolean` to the `DailyMission` interface in `useDailies.ts`
- `GET /api/dailies` now includes `desktopOnly` on each mission object in the response
- `DailiesWidget.tsx` shows a **"PC only"** badge next to any `desktopOnly` mission when `isMobile = true`, with a tooltip: *"This mission can only be completed on a PC, not on a phone."*
- The FLY shortcut button is also hidden on mobile since fly missions are desktop-only anyway

This means mobile users will see all 3 missions assigned to them (same as desktop), but `desktopOnly` missions will clearly indicate they can't be completed on mobile — no confusion, no broken progress tracking.

Ixotic27:
Hi @hariom888, thanks for the PR! Instead of just adding a 'PC only' badge and hiding the fly button, could you please add a message/alert when the user tries to do the flying task on their phone, explicitly saying that 'this mission can be done on the pc not on phone'? Also, please provide screenshots of this message while accessing that task so that the styling, layout, and working of the message can be checked. A video will be much preferred!

hariom888:
Hi @Ixotic27! Here are the screenshots of the PC-only alert on mobile view.

**1. Mission list showing the "PC only" badge next to the Casual Pilot (fly) mission:**

image

**2. Alert modal that appears when the user taps the "PC only" badge:**

image

The modal dismisses on "Got it" or by tapping outside. The FLY button is also hidden on mobile.

Ixotic27 (approved):

PR #225, opened 2026-06-03
fix: add atomic execute_raid RPC to replace non-atomic fallback path 

### LINKED ISSUE #206: [Bug]: Raid fallback execution path is non-atomic and has no rollback on partial failure
### Bug Description

`POST /api/raid/execute` contains a fallback path triggered when the `execute_raid` Postgres RPC is not found (error message includes `"execute_raid"`). This fallback performs the full raid side-effects as sequential, unguarded `await` calls with no database transaction and no rollback logic:

1. Insert raid row
2. Update `last_raided_at` + `active_defenses` on defender
3. Consume attacker boost (`purchases.status = 'consumed'`)
4. Consume attacker consumable via `consumeDeveloperItem`
5. Consume defender consumable via `consumeDeveloperItem`
6. Update `raid_xp` on both attacker and defender
7. Call `grant_xp` RPC twice
8. Insert into `raid_tags`
9. Insert into `activity_feed`

A transient error or timeout after step 1 but before any subsequent step leaves the database in an inconsistent state: the raid is recorded, but the peace shield may not be applied, items are not consumed, and XP is not granted. A retry of the same request will no-op on the raid row insert (duplicate guard) but silently skip all remaining side-effects.

This fallback is the current live path for any deployment that has not run the `execute_raid` RPC migration.

**Affected file:** `src/app/api/raid/execute/route.ts`, lines inside `if (raidError?.message?.includes("execute_raid"))`.

### Steps to Reproduce

1. **Short-term:** Promote the `execute_raid` Postgres function to required (document migration step clearly), and have the fallback path return a `503` rather than silently proceeding with non-atomic writes.
2. **Long-term / preferred:** Rewrite the fallback to use a single Postgres function that wraps all side-effects in a `BEGIN/COMMIT` block, returning a structured result. The existing pattern of `execute_raid` is already the right direction — it just needs to be shipped as a required migration.

### Expected Behavior

The raid is recorded, but the peace shield may not be applied, items are not consumed, and XP is not granted

_No response_

### Browser

_No response_

### Acknowledgements

Srejoye:
## What does this PR do?

The fallback path in `src/app/api/raid/execute/route.ts` (triggered when the `execute_raid` RPC was missing) performed 9+ sequential database mutations with no transaction and no rollback. A failure at any step after the raid row was inserted left the database in a partial state — peace shield not applied, items not consumed, XP not granted.

The `execute_raid` RPC did not exist anywhere in the migrations, meaning the fallback was the live path for all deployments.

Added `supabase/migrations/047_execute_raid_atomic.sql` which defines the `execute_raid` Postgres function, wrapping all raid side-effects in a single atomic transaction — raid row insert, peace shield reset, tag updates, raid_xp grants, and activity feed insert. With this migration applied, the RPC path succeeds and the non-atomic fallback is never reached.

## Visual Proof

This is a database-level atomicity fix with no UI changes to screenshot.

**Before — 9 unguarded sequential awaits in the fallback (non-atomic):**
```ts
await admin.from("raids").insert({ ... })           // step 1
await admin.from("developers").update({ ... })      // step 2 — if this fails, raid row is orphaned
await admin.from("purchases").update({ ... })       // step 3
await consumeDeveloperItem(attacker.id, ...)        // step 4
await consumeDeveloperItem(defender.id, ...)        // step 5
await admin.from("developers").update({ raid_xp })  // step 6
await admin.rpc("grant_xp", { attacker })           // step 7
await admin.rpc("grant_xp", { defender })           // step 8
await admin.from("activity_feed").insert({ ... })   // step 9
// ❌ any failure between steps leaves DB in partial state
```

**After — single atomic Postgres function:**
```sql
-- All side-effects inside one implicit transaction
-- If any statement fails, the entire function rolls back
INSERT INTO raids ...              -- atomic
UPDATE developers SET last_raided_at ...  -- atomic
UPDATE raid_tags ...               -- atomic
UPDATE developers SET raid_xp ... -- atomic
INSERT INTO activity_feed ...      -- atomic
-- ✅ guaranteed all-or-nothing
```

## Related issue
Fixes #206

## Checklist

### DIFF
--- a/supabase/migrations/047_execute_raid_atomic.sql
+++ b/supabase/migrations/047_execute_raid_atomic.sql
@@ -1,0 +1,107 @@
+CREATE OR REPLACE FUNCTION execute_raid(
+  p_attacker_id       bigint,
+  p_defender_id       bigint,
+  p_attack_score      int,
+  p_defense_score     int,
+  p_success           boolean,
+  p_attack_breakdown  jsonb,
+  p_defense_breakdown jsonb,
+  p_vehicle           text,
+  p_tag_style         text
+)
+RETURNS jsonb LANGUAGE plpgsql AS $$
+DECLARE
+  v_raid_id           bigint;
+  v_attacker_raid_xp  int;
+  v_defender_raid_xp  int;
+BEGIN
+  -- Prevent duplicate raids in the same request (idempotency guard)
+  IF EXISTS (
+    SELECT 1 FROM raids
+    WHERE attacker_id = p_attacker_id
+      AND defender_id = p_defender_id
+      AND created_at >= now() - interval '10 seconds'
+  ) THEN
+    RAISE EXCEPTION 'duplicate_raid';
+  END IF;
+
+  -- 1. Insert raid row
+  INSERT INTO raids (
+    attacker_id, defender_id,
+    attack_score, defense_score,
+    success,
+    attack_breakdown, defense_breakdown,
+    attacker_vehicle, attacker_tag_style
+  )
+  VALUES (
+    p_attacker_id, p_defender_id,
+    p_attack_score, p_defense_score,
+    p_success,
+    p_attack_breakdown, p_defense_breakdown,
+    p_vehicle, p_tag_style
+  )
+  RETURNING id INTO v_raid_id;
+
+  -- 2. Apply peace shield + clear active defenses on defender
+  UPDATE developers
+  SET last_raided_at = now(),
+      active_defenses = '[]'::jsonb
+  WHERE id = p_defender_id;
+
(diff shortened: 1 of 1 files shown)


Srejoye:
Hey @Ixotic27 👋

This is a database-level atomicity fix — all side-effects happen inside a single Postgres function which rolls back automatically if any statement fails, so there's no UI behaviour change to record as a screenshot or GIF.

I've added a before/after code diff in the PR description showing the non-atomic fallback (9 sequential awaits with no rollback) vs the new atomic RPC. The fix is fully verifiable from the migration file in this PR.

Happy to provide anything else you need for review!

PR #233, opened 2026-06-03
fix: sanitize LED banner text to strip control/bidi/HTML characters

### LINKED ISSUE #231: [Bug]: LED Banner customization text not sanitized — potential stored XSS via building display
### Bug Description

In `src/app/api/customizations/route.ts` (line 209), the LED banner text is stored with only a length limit (100 chars) but **no sanitization**:

```ts
config: { text: text.substring(0, 100) }
```

The stored text is then rendered directly in the 3D city via `Building3D.tsx` as part of the LED banner overlay. If the rendering path uses `dangerouslySetInnerHTML` or similar unsanitized output, a user could inject malicious HTML/JS that renders on every visitor's browser when they view that building.

Even if React escapes the output by default, the text could contain:
- Unicode control characters that break layout
- Zero-width characters that appear empty but carry hidden content
- Emoji/symbol floods that cause rendering issues

### Steps to Reproduce
1. Own the LED Banner item
2. Set text to a string containing control characters or extremely long unicode sequences
3. Observe the text renders without sanitization

### Expected Behavior
The text should be sanitized to:
1. Strip HTML tags
2. Remove control characters (Unicode categories Cc, Cf)
3. Trim whitespace
4. Validate the string is renderable plain text

### Suggested Fix
```ts
const sanitized = text
  .replace(/[\u0000-\u001F\u007F-\u009F]/g, '') // strip control chars
  .replace(/<[^>]*>/g, '')  // strip HTML tags
  .trim()
  .substring(0, 100);
```

### Files
- `src/app/api/customizations/route.ts` (line 209)

Srejoye:
## What does this PR do?

Fixes unsanitized storage of LED banner text in `src/app/api/customizations/route.ts`. Previously, the only guard was `.substring(0, 100)`, which allowed control characters, bidirectional override sequences, zero-width characters, and HTML tag fragments to reach the database and be rendered in the 3D city scene.

**Note on the actual attack surface:** after tracing the full rendering path (`route.ts` → `Building3D.tsx` → `BuildingEffects.tsx` → `@react-three/drei <Text>`), the output is a WebGL canvas draw, not a DOM insertion. Classical stored XSS via script injection is therefore not exploitable. However, the following remain valid and are fixed by this PR:

- **Bidirectional override characters** (e.g. U+202E) — can reverse rendered text and confuse moderation tooling
- **Zero-width characters** — produce strings that appear empty in the UI but aren't, bypassing client-side length validation
- **C0/C1 control characters** — can corrupt the MSDF font atlas used by `drei <Text>` and cause unpredictable rendering for all visitors of that building tile
- **HTML tag fragments** — belt-and-suspenders; irrelevant today but prevents breakage if the rendering path ever gains a DOM fallback

## Files changed

- `src/lib/sanitize-led-banner.ts` — new pure helper function `sanitizeLedBannerText`
- `src/app/api/customizations/route.ts` — import and apply the helper in the `led_banner` branch; both the upserted value and the JSON response now use the sanitized string; a string that reduces to empty after sanitization triggers the existing delete path (consistent with an explicit `null` clear)
- `src/lib/__tests__/sanitize-led-banner.test.ts` — 25 unit tests covering happy path, emoji/unicode passthrough, length enforcement (truncation applied after stripping), each stripped character category individually, and null-equivalent edge cases

## Testing

```bash
pnpm test src/lib/__tests__/sanitize-led-banner.test.ts
```

All 25 cases pass. No changes to existing tests.

## Checklist

## Related issue

Fixes #231


Ixotic27 (approved):
✅ Excellent, secure sanitization implementation for LED Banner text with a comprehensive test suite! Code is robust and ready to merge.

Issue #230, opened 2026-06-03
[Bug]: Kudos daily limit uses local server timezone instead of UTC, causing inconsistent limits


Ixotic27:
### Bug Description

In `src/app/api/interactions/kudos/route.ts` (line 65), the daily date is computed using the server's local timezone:

```ts
const today = new Date().toISOString().split("T")[0];
```

While `toISOString()` outputs UTC, the `given_date` column in `developer_kudos` may contain dates in different timezone representations depending on when other parts of the system write to it.

More critically, lines 115-116 compute "yesterday" for streak tracking:
```ts
const yesterday = new Date(Date.now() - 86400000).toISOString().split("T")[0];
```

This hardcodes exactly 24 hours, which breaks during DST transitions (if the server runs in a DST-aware timezone) and doesn't account for UTC day boundaries consistently.

### Steps to Reproduce
1. Give kudos at 23:55 UTC
2. Give kudos again at 00:05 UTC (next day)
3. The 24-hour subtraction for "yesterday" check may miscount streak days around DST changes

### Expected Behavior
All date calculations should use explicit UTC methods (`new Date().toISOString().split('T')[0]` is correct for the date, but the streak calculation at line 115 should also be ISO-based).

### Suggested Fix
Replace the `yesterday` calculation with a proper UTC date subtraction:
```ts
const todayDate = new Date();
const yesterdayDate = new Date(Date.UTC(todayDate.getUTCFullYear(), todayDate.getUTCMonth(), todayDate.getUTCDate() - 1));
const yesterday = yesterdayDate.toISOString().split("T")[0];
```

### Files
- `src/app/api/interactions/kudos/route.ts` (lines 65, 115-116)

diksha78dev:
assign me!

### Proposed Solution

I will fix the UTC date handling in `src/app/api/interactions/kudos/route.ts` to ensure consistent daily limits and streak tracking.

* Review the current date calculations used for kudos limits and streak validation.
* Replace the fixed `24-hour` subtraction logic with a proper UTC-based date calculation.
* Ensure all day comparisons use UTC date boundaries consistently.
* Test edge cases around day transitions and timezone/DST changes to verify correct streak behavior.

This will prevent inconsistencies caused by timezone differences and make kudos tracking reliable across all environments.

Srejoye:
assign me

### Proposed Solution

```ts
const yesterday = new Date(Date.now() - 86400000).toISOString().split("T")[0];
```

[...]

PR #253, opened 2026-06-04
fix: atomic first-solve guard prevents duplicate XP/points/item grants on concurrent submissions

### LINKED ISSUE #238: [Bug]: Arena submission rewards granted via read-then-write race — users can claim first-solve XP, points, and item drops multiple times
### Bug Description

`POST /api/arena/submit` (`src/app/api/arena/submit/route.ts`) determines whether a submission is a first-solve using a read-then-write pattern with no database-level guard:

1. The submission row is **inserted** (line 146–160).
2. A second query reads back accepted submissions for the same challenge/problem and counts them (lines 177–193).
3. If `prior.length > 1`, it's not a first solve.

Because step 1 completes before step 2 reads, any two concurrent requests that both pass step 1 before either reads step 2 will **both** see `prior.length === 1` (only their own row), both set `isFirstSolve = true`, and both receive:
- Arena points increment (`developers.points += grantedPoints`, lines 230–234)
- XP via `grant_xp` RPC (line 237)
- Random item drops written to `arena_inventory` (lines 72–95)
- Rating and streak increments in `arena_ratings` (lines 289–298)

The points increment is a read-modify-write (fetch current points → add → update) with no optimistic lock or atomic RPC, so two concurrent winners each read the same baseline and both write `baseline + grantedPoints`, making one grant invisible.

This is trivially reproducible by submitting the same accepted solution twice in rapid succession from two browser tabs or with two parallel curl requests. There is no idempotency key on the submission, no unique constraint preventing duplicate accepted rows for the same `(user_id, challenge_id)`, and no transaction wrapping the read-check + reward grant.

### Steps to Reproduce

1. Obtain a valid accepted solution for any arena challenge.
2. Fire two concurrent `POST /api/arena/submit` requests with `status: "accepted"` and identical `challenge_id`.
3. Both requests insert a submission row, then both read exactly 1 accepted row (their own), both conclude `isFirstSolve = true`.
4. Both grant XP, points, and item drops to the same user.

### Expected Behavior

First-solve rewards should be granted exactly once per `(user_id, challenge_id)`. Duplicate grants must be prevented at the database level, not in application logic.

_No response_

### Browser

_No response_

### Acknowledgements

Srejoye:
## What does this PR do?

Fixes a read-then-write race in `POST /api/arena/submit` that allowed two concurrent accepted submissions for the same challenge to both receive first-solve rewards (XP, points, item drops).

**The race (before):**
1. Request A inserts submission row
2. Request B inserts submission row
3. Both query back `prior.length` — both see `1` (only their own row)
4. Both set `isFirstSolve = true` and both execute the full reward path

The `developers.points` increment compounded this: both requests read the same baseline and write `baseline + grantedPoints`, making one grant invisible.

**The fix:**

A new `arena_first_solves` table with `UNIQUE(user_id, challenge_id)` and a partial unique index on `(user_id, problem_id)` acts as the database-level guard. The new `claim_first_solve()` Postgres RPC does:

```sql
INSERT INTO arena_first_solves ... ON CONFLICT DO NOTHING;
-- returns won_race = (ROW_COUNT > 0)
-- if won_race: UPDATE developers SET points = points + $amount
```

Only the first concurrent caller gets `won_race = true`. All others (including retries) get `false` and skip rewards entirely. Points
increment is an atomic `UPDATE ... SET points = points + $1` — no fetch-then-write.

Item drops also had a read-then-write in `rollItemDrops`; replaced with a new `upsert_arena_inventory_item()` RPC that uses `INSERT ... ON CONFLICT DO UPDATE SET quantity = quantity + 1`.

**Bonus fix:** The streak `yesterdayStr` calculation was using `Date.now() - 86_400_000ms` (DST-sensitive); replaced with `Date.UTC(y, m, d - 1)` consistent with the fix in #230. 

## Files changed

- `supabase/migrations/049_arena_first_solve_idempotency.sql` — new table, 3 RPCs
- `src/app/api/arena/submit/route.ts` — replace prior-count check + fetch-then-write points with RPC calls
- `src/lib/__tests__/arena-first-solve.test.ts` — 14 unit tests

## Testing

```bash
pnpm test src/lib/__tests__/arena-first-solve.test.ts
```

## Checklist

## Related issue

Fixes #238

### DIFF
--- a/src/app/api/arena/submit/route.ts
+++ b/src/app/api/arena/submit/route.ts
@@ -20,46 +20,33 @@ async function rollItemDrops(sb: any, difficulty: string, devId: number): Promis
   const roll = Math.random();
 
   if (difficulty === "easy") {
-    // 100% -> 1 Common item
     const common = await getRandomItemByRarity("common");
     if (common) droppedItems.push(common);
-
-    // 15% -> 1 Rare item
     if (roll < 0.15) {
       const rare = await getRandomItemByRarity("rare");
       if (rare) droppedItems.push(rare);
     }
   } else if (difficulty === "medium") {
-    // 100% -> 1 Rare item
     const rare = await getRandomItemByRarity("rare");
     if (rare) droppedItems.push(rare);
-
-    // 20% -> 1 Epic item
     if (roll < 0.20) {
       const epic = await getRandomItemByRarity("epic");
       if (epic) droppedItems.push(epic);
-    }
-    // 10% -> 2nd Rare item
-    else if (roll < 0.30) {
+    } else if (roll < 0.30) {
       const rare2 = await getRandomItemByRarity("rare");
       if (rare2) droppedItems.push(rare2);
     }
   } else if (difficulty === "hard") {
-    // 100% -> 1 Epic item
     const epic = await getRandomItemByRarity("epic");
     if (epic) droppedItems.push(epic);
-
-    // 25% -> 1 Rare item
     if (roll < 0.25) {
       const rare = await getRandomItemByRarity("rare");
       if (rare) droppedItems.push(rare);
     }
-    // 5% -> 1 Legendary item
     if (Math.random() < 0.05) {
       const legendary = await getRandomItemByRarity("legendary");
       if (legendary) droppedItems.push(legendary);
     }
-    // 8% -> Epic + Rare bonus combo
     if (Math.random() < 0.08) {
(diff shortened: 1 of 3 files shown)


Srejoye:
@Ixotic27 

## Visual Proof

This is a **backend race condition fix** with no UI changes. The bug only manifests under concurrent requests hitting the same endpoint simultaneously.

Ixotic27 (approved):

PR #254, opened 2026-06-04
fix: enforce raid cooldown, daily cap and peace-shield atomically in DB — removes serverless-bypassable in-process rate limiter

### LINKED ISSUE #240: [Bug]: Raid daily cap and peace-shield checks use in-process rate limiter + pre-insert reads, allowing bypass across serverless instances
### Bug Description

`POST /api/raid/execute` (`src/app/api/raid/execute/route.ts`) enforces three sequential guards before inserting a raid:

1. **In-memory rate limit** (`rateLimit('raid-execute:{user.id}', 1, 30_000)`) — line 66
2. **Daily cap check** — a `COUNT` query against `raids` for `attacker_id` today (lines 117–131)
3. **2-hour peace shield** — reads `defender.last_raided_at` fetched at line 91 (before any write) and compares to `new Date()` (lines 134–138)
4. **Weekly pair cooldown** — a second `COUNT` query (lines 144–158)

All three guards are evaluated **before** the `execute_raid` RPC or direct insert (line 371 / lines 386–400). This creates two distinct failure modes:

**Failure mode A — Serverless bypass of the in-memory rate limiter**

`src/lib/rate-limit.ts` explicitly documents: *"State is per-process: each serverless cold-start gets its own Map."* On Vercel, two requests arriving on two cold-started instances share no rate-limit state. A user who fires two simultaneous raid requests will hit two separate Node.js processes with empty stores, both returning `ok: true`. The 30-second cooldown is fully bypassable.

**Failure mode B — Read-then-insert race on daily cap and peace shield**

Both the daily cap count (line 121) and the peace-shield timestamp (from the pre-fetch at line 91, before any write) are read before the raid is committed. Two concurrent raids from the same attacker targeting different defenders both pass the daily cap check if they arrive before either's insert completes. Similarly, two attackers targeting the same defender can both read a stale `last_raided_at` and both bypass the peace shield if they race between the shield read and the `last_raided_at` update (line 412).

The fallback path (lines 383–600, active when `execute_raid` RPC is missing) applies the peace shield update as a **separate** un-guarded `UPDATE` after the insert, widening the window.

### Steps to Reproduce

**Failure mode A:**
1. Open two browser tabs or use two parallel processes.
2. Fire `POST /api/raid/execute` simultaneously from both; each will land on a cold Vercel instance.
3. Both pass the 30-second rate limit because the in-memory store is per-process.

**Failure mode B (peace shield):**
1. Two users (A and B) both target user C simultaneously.
2. Both fetch `defender.last_raided_at` before either inserts.
3. Both see no active shield, both insert successfully, both update `last_raided_at`.
4. C receives two simultaneous raids in violation of the 2-hour shield.

### Expected Behavior

The 30-second cooldown, daily cap, and peace shield must be enforced atomically at the database layer, not in application memory or via separate pre-reads.

_No response_

### Browser

_No response_

### Acknowledgements

Srejoye:
## What does this PR do?

Fixes two failure modes in `POST /api/raid/execute` that allowed concurrent requests to bypass the 30-second cooldown, daily cap, and 2-hour peace shield.

**Failure mode A — serverless bypass of in-process rate limiter**

`src/lib/rate-limit.ts` uses a module-level `Map`. Its own docstring says *"State is per-process."* On Vercel, two simultaneous requests land on two cold-started instances with empty stores — both return `ok: true`. The 30-second cooldown was fully bypassable. The `rateLimit()` call is removed from this route entirely.

**Failure mode B — read-then-insert race**

Daily cap (`COUNT` at line 117) and peace shield (`last_raided_at` read at line 91) were both fetched before any INSERT. Two concurrent requests both read stale values and both passed the guards.

**The fix — `execute_raid()` RPC**

All four guards (cooldown, daily cap, peace shield, weekly pair) now execute atomically inside a single `plpgsql` function alongside the `INSERT`:

- **Cooldown**: `INSERT ... ON CONFLICT DO UPDATE WHERE cooldown_until <= now()` — returns `ROW_COUNT = 0` for the second concurrent caller, blocking it instantly.
- **Peace shield**: `SELECT ... FOR UPDATE` locks the defender row so two attackers targeting the same user serialize correctly.
- **Daily cap / weekly pair**: `COUNT` queries run inside the same transaction after the cooldown CAS succeeds.

The JS fallback path (lines 383–601) is removed. `execute_raid()` is now the single authoritative code path.

## Files changed

- `supabase/migrations/050_raid_atomic_guards.sql` — `raid_cooldowns` table + `execute_raid()` RPC
- `src/app/api/raid/execute/route.ts` — remove `rateLimit()` call, remove pre-insert guard reads, remove JS fallback path, map RPC result to HTTP responses
- `src/lib/__tests__/raid-atomic-guards.test.ts` — 16 unit tests

## Visual Proof

This is a **backend concurrency fix** with no UI changes. The bugs only manifest under simultaneous requests. 

## Testing

```bash
npx vitest run src/lib/__tests__/raid-atomic-guards.test.ts
```

## Checklist

## Related issue

Fixes #240

### DIFF
--- a/src/app/api/raid/execute/route.ts
+++ b/src/app/api/raid/execute/route.ts
@@ -1,7 +1,6 @@
 import { NextResponse } from "next/server";
 import { createServerSupabase } from "@/lib/supabase-server";
 import { getSupabaseAdmin } from "@/lib/supabase";
-import { rateLimit } from "@/lib/rate-limit";
 import { checkAchievements } from "@/lib/achievements";
 import { touchLastActive } from "@/lib/notification-helpers";
 import { sendRaidAlertNotification } from "@/lib/notification-senders/raid";
@@ -10,7 +9,6 @@ import {
   calculateAttackScore,
   calculateDefenseScore,
   getRaidTitle,
-  MAX_RAIDS_PER_DAY,
   RAID_TAG_DURATION_DAYS,
   XP_WIN_ATTACKER,
   XP_WIN_DEFENDER,
@@ -62,11 +60,11 @@ export async function POST(request: Request) {
     return NextResponse.json({ error: "Not authenticated" }, { status: 401 });
   }
 
-  // Strict per-user rate limit: 1 execute per 30s
-  const { ok } = rateLimit(`raid-execute:${user.id}`, 1, 30_000);
-  if (!ok) {
-    return NextResponse.json({ error: "Too fast, wait before raiding again" }, { status: 429 });
-  }
+  // NOTE: The in-process rateLimit() call is intentionally removed.
+  // It used a per-process Map (src/lib/rate-limit.ts) with no shared
+  // state across serverless instances — trivially bypassable on Vercel.
+  // The 30-second cooldown is now enforced atomically inside execute_raid()
+  // via the raid_cooldowns table CAS pattern.
 
   const body = await request.json();
   const { target_login, boost_purchase_id, consumable_item_id: _legacy_consumable, offensive_item_id, vehicle_id } = body as {
@@ -76,7 +74,6 @@ export async function POST(request: Request) {
     offensive_item_id?: string;
     vehicle_id?: string;
   };
-  // Support both the old `consumable_item_id` key and the new `offensive_item_id` key
   const consumable_item_id = offensive_item_id ?? _legacy_consumable;
 
   if (!target_login || typeof target_login !== "string") {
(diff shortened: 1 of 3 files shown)


Srejoye:
@Ixotic27 This is a backend concurrency fix with no UI changes. The bugs only manifest under simultaneous requests.

Ixotic27 (approved):
This is a great and robust fix for the concurrency issues! Thank you for the atomic DB transaction approach, it's exactly what was needed and ensures absolute correctness under load. @Srejoye, if you're interested, we have about 4 other open issues right now — please feel free to check them out and request to work on any of them!

PR #274, opened 2026-06-04
fix: atomic XP code redemption — prevent double-grant race, stale used_count, and XP-before-usage ordering

### LINKED ISSUE #258: [Bug]: XP Redeem Code — Non-Atomic Usage Tracking Allows Double-Redemption Under Concurrency
## Description

In `src/app/api/shop/redeem-xp/route.ts`, the XP code redemption flow has a classic read-then-write race condition:

1. **Line 59-64**: Check if user already redeemed this code (`xp_code_usages` SELECT)
2. **Line 75-78**: Apply XP to developer (UPDATE `developers`)
3. **Line 85-88**: Record the usage (INSERT `xp_code_usages`)
4. **Line 97-99**: Increment `used_count` (UPDATE `xp_redeem_codes`)

Two concurrent requests from the same user can both pass step 1 (no usage row exists yet), both apply XP in step 2, and both insert usage rows in step 3. This allows a user to redeem the same code twice, getting double XP.

**Additionally**, the `used_count` increment at line 97 uses a stale snapshot value (`redeemCode.used_count + 1`) instead of an atomic DB increment, so concurrent requests from different users can also cause the count to drift.

## Expected Behavior
- Each user can only redeem a code once
- `used_count` increments atomically

## Suggested Fix
- Use an `INSERT ... ON CONFLICT DO NOTHING` on `xp_code_usages` and check the returned row to determine if the insert actually happened (similar to how kudos handles duplicates)
- Use an atomic SQL increment for `used_count` instead of a JS snapshot value

Srejoye:
## What does this PR do?

Fixes three independent bugs in `POST /api/shop/redeem-xp` that together allow a user to redeem the same code twice and receive double XP under concurrent load.

**Bug 1 — Double-redemption race**

The route did `SELECT xp_code_usages` (line 59) then `INSERT` (line 85) with no DB-level uniqueness guard. Two concurrent requests both read `existingUsage = null`, both apply XP, both insert. Fixed by adding `UNIQUE(code_id, developer_id)` to `xp_code_usages` and replacing the SELECT+INSERT with `INSERT ... ON CONFLICT DO NOTHING` inside the `redeem_xp_code()` RPC — only the first concurrent caller gets `ROW_COUNT = 1`; all others are blocked.

**Bug 2 — Stale-snapshot `used_count` increment**

Lines 91–99 read `redeemCode.used_count` into a JS variable, added 1, and wrote it back. Two concurrent users targeting a code with `max_uses = 5` both read `used_count = 4`, both wrote `5`, silently bypassing the cap. Fixed with `UPDATE SET used_count = used_count + 1 WHERE used_count < max_uses` inside the RPC — if 0 rows updated, the usage insert is rolled back.

**Bug 3 — XP applied before usage recorded**

The XP `UPDATE` (line 75) fired before the `xp_code_usages` `INSERT` (line 85). An `INSERT` failure left the user with XP and no usage record, allowing re-redemption. Fixed by reordering: usage insert → used_count increment → XP update, with XP only applied after RPC returns `ok = true`.

## Files changed

- `supabase/migrations/051_xp_redeem_codes_idempotency.sql` — unique constraint + `redeem_xp_code()` RPC
- `src/app/api/shop/redeem-xp/route.ts` — replace SELECT+INSERT pattern with RPC call; reorder XP update after confirmation
- `src/lib/__tests__/xp-redeem.test.ts` — 17 unit tests

## Visual Proof

⚠️ This is a **backend race condition fix** with no UI changes.

## Testing

```bash
npx vitest run src/lib/__tests__/xp-redeem.test.ts
```

## Checklist

## Related issue

Fixes #258

### DIFF
--- a/src/app/api/shop/redeem-xp/route.ts
+++ b/src/app/api/shop/redeem-xp/route.ts
@@ -20,8 +20,8 @@ export async function POST(req: Request) {
     }
 
     const sb = getSupabaseAdmin();
-    
-    // First, verify the user has a linked developer account
+
+    // Verify the user has a linked developer account
     const { data: dev } = await sb
       .from("developers")
       .select("id, xp_total")
@@ -29,13 +29,16 @@ export async function POST(req: Request) {
       .single();
 
     if (!dev) {
-      return NextResponse.json({ error: "You must link a LeetCode account first." }, { status: 403 });
+      return NextResponse.json(
+        { error: "You must link a LeetCode account first." },
+        { status: 403 }
+      );
     }
 
-    // Attempt to find the code in the xp_redeem_codes table
+    // Fetch and validate the code
     const { data: redeemCode, error: fetchError } = await sb
       .from("xp_redeem_codes")
-      .select("*")
+      .select("id, xp_amount, max_uses, used_count, expires_at")
       .eq("code", code.trim().toUpperCase())
       .single();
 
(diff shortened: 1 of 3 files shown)


Ixotic27 (approved):

PR #275, opened 2026-06-04
fix(checkin): idempotent side-effects for concurrent serverless instances

### LINKED ISSUE #259: [Bug]: Checkin Route Uses In-Process Rate Limiter — Bypassable on Serverless
## Description

In `src/app/api/checkin/route.ts` (line 104), the check-in endpoint uses the in-process `rateLimit()` from `src/lib/rate-limit.ts`. As documented in PR #254, this rate limiter uses a **per-process `Map`** with no shared state across serverless instances.

On Vercel, two simultaneous check-in requests can land on two different cold-started instances, both with empty rate-limit stores — both will return `ok: true`. While the `perform_checkin` RPC likely handles idempotency for the actual streak logic, the surrounding side effects (XP grants via `grant_xp`, achievement checks, activity feed inserts, daily mission tracking) could execute multiple times.

## Affected Endpoints
The same vulnerability applies to ALL routes using the in-process `rateLimit()`:
- `/api/checkin` — streak XP, achievements, feed events
- `/api/interactions/kudos` — daily cap bypass (see issue #257)
- `/api/fly-scores` — unlimited XP farming (see issue #256)
- `/api/rabbit` — quest progress bypass
- `/api/shop/buy-with-points` — purchase rate limit
- `/api/interactions/visit` — visit tracking

## Expected Behavior
Rate limiting should be enforced atomically at the database level (e.g., via a cooldown table like `raid_cooldowns` from PR #254) or via a shared store (Redis/Upstash).

## Suggested Fix
Either:
1. Move critical rate limits to DB-level atomic checks (like the raid cooldown CAS pattern)
2. Use Upstash Redis for shared rate limiting across serverless instances
3. At minimum, ensure the side effects are idempotent so duplicate executions are harmless

Srejoye:
## What does this PR do? 

The in-process `rateLimit()` in `src/lib/rate-limit.ts` holds state in a per-process `Map`. On Vercel, two simultaneous check-in requests can land on two different cold-started instances, both with empty stores — both pass the rate-limit guard and race through the side-effects.

This PR does **not** replace the rate limiter (that's a larger, separate change). Instead it makes every side-effect after `perform_checkin` returns idempotent at the DB level, so concurrent executions are harmless.

## Risk audit

| Side-effect | Was it safe? | Fix |
|---|---|---|
| `perform_checkin` RPC | ✅ `last_checkin_date` guard in DB | No change |
| `grantStreakReward` | ✅ `upsert ... ignoreDuplicates` on `streak_rewards` | No change |
| `trackDailyMission` | ✅ upsert-based internally | No change |
| `grant_xp` | ❌ fires unconditionally | Gated by `checkin_xp_log` unique insert |
| `activity_feed.insert` | ❌ plain INSERT, duplicates on race | `upsert` on `(actor_id, event_type, event_date)` |
| `streak_freeze_log.insert` | ❌ plain INSERT | `upsert` on `(developer_id, action, granted_date)` |

## Changes

- **`src/app/api/checkin/route.ts`** — three targeted changes (XP log guard, feed upsert, freeze log upsert); no control-flow restructuring
- **`supabase/migrations/20260604_checkin_idempotency.sql`** — creates `checkin_xp_log`; adds `event_date` + unique constraint to `activity_feed`; adds `granted_date` + unique constraint to `streak_freeze_log`. All `ALTER TABLE` steps are wrapped in `DO $$` existence checks — safe to re-run.
- **`src/lib/__tests__/checkin-idempotency.test.ts`** — unit tests for the 23505 guard logic and a documented regression test showing why the in-process limiter alone cannot prevent the race.

## Testing

```bash
npm test src/lib/__tests__/checkin-idempotency.test.ts
```

Apply the migration against your local Supabase instance:
```bash
supabase db reset
# or
supabase migration up
```

## Before / after

**Before:** two concurrent check-in requests on cold instances → two `grant_xp` calls (+20 XP instead of +10), two `activity_feed` rows for the same event, possible duplicate `streak_freeze_log` rows.

**After:** only the first INSERT into `checkin_xp_log` succeeds; the second hits a `23505` unique violation and skips XP silently. Feed and freeze log are upserted — second write is a no-op.

## Checklist

## Related issue
Fixes #259

### DIFF
--- a/src/app/api/checkin/route.ts
+++ b/src/app/api/checkin/route.ts
@@ -179,10 +179,30 @@ export async function POST(request: Request) {
   let streakReward: { milestone: number; item_id: string; item_name: string } | null = null;
   let xpResult: { granted: number; new_total: number; new_level: number } | null = null;
 
-  // Grant XP for check-in
+  // Grant XP for check-in — idempotent via unique (developer_id, source, date) insert.
+  // If two cold instances both reach here, only the first INSERT wins; the second
+  // sees a 23505 conflict and skips the grant entirely.
   if (checkinResult.checked_in) {
-    const { data: xpData } = await sb.rpc("grant_xp", { p_developer_id: dev.id, p_source: "checkin", p_amount: 10 });
-    if (xpData) xpResult = xpData as { granted: number; new_total: number; new_level: number };
+    const today = new Date().toISOString().split("T")[0];
+    const { error: xpLogError } = await sb
+      .from("checkin_xp_log")
+      .insert({ developer_id: dev.id, granted_date: today })
+      .select("id")
+      .maybeSingle();
+
+    if (!xpLogError) {
+      // Insert succeeded — we are the first instance, safe to grant XP
+      const { data: xpData } = await sb.rpc("grant_xp", {
+        p_developer_id: dev.id,
+        p_source: "checkin",
+        p_amount: 10,
+      });
+      if (xpData) xpResult = xpData as { granted: number; new_total: number; new_level: number };
+    } else if (!xpLogError.code?.includes("23505")) {
+      // Unexpected error (not a duplicate) — log but don't crash
+      console.error("[checkin] checkin_xp_log insert error:", xpLogError);
+    }
+    // 23505 = duplicate, another instance already granted XP today — skip silently
   }
 
   if (checkinResult.checked_in) {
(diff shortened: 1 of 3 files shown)


Ixotic27 (approved):

PR #281, opened 2026-06-04
fix(#255): prevent TOCTOU race in redeem-special all_items grant

### LINKED ISSUE #255: [Bug]: Redeem-Special "all_items" Race Condition — Infinite Item Duplication
## Summary
The "all_items" code type handler has a TOCTOU race condition: purchases are INSERTed *before* the usage record is created, and there is no transaction or locking. Concurrent requests can each pass the `existingUsage` check and duplicate every item in the shop.

## Root Cause
```typescript
// Step order:
1. SELECT special_code_usages → null         (both requests see null)
2. SELECT owned purchases → []                (both see empty)
3. INSERT purchases for all items             (BOTH succeed)
4. INSERT special_code_usages                 (first succeeds, second silently fails on PK)
5. UPDATE used_count
```

Step 3 happens before step 4, and steps 3–5 are not wrapped in a DB transaction. The Supabase client does not throw on constraint violations — `.insert()` returns an error object that is never inspected.

## Impact
Infinite duplication of any item in the shop. Complete destruction of the game economy — users can obtain all shop items for free repeatedly with no audit trail.

## Suggested Fix
Wrap the entire grant flow in a Supabase RPC that atomically checks usage, inserts purchases, and records usage. Alternatively:

1. Add a unique constraint on `(developer_id, item_id)` in `purchases` with `ON CONFLICT DO NOTHING`
2. Insert the usage record BEFORE purchase inserts, checking its error:

```sql
ALTER TABLE purchases ADD UNIQUE (developer_id, item_id);
```

### Acknowledgements

ionfwsrijan:
## Summary
Fixes a TOCTOU race condition in the "all_items" code type handler that allowed infinite item duplication via concurrent requests.

## Root Cause
Purchases were INSERTed **before** the usage record was created, with no transaction or locking. Two concurrent requests both pass the `existingUsage` SELECT check (seeing null), then both insert purchases for all shop items. The usage insert (step 4) fails silently on PK constraint for the second request, but the damage is done — both purchase batches succeed.

### `src/app/api/shop/redeem-special/route.ts`
- **Reordered operations**: Usage record (`special_code_usages`) is now inserted **before** purchases, making the unique constraint on `(code_id, developer_id)` an atomic gate.
- **Added error check**: If the usage insert fails with `23505` (unique violation), the endpoint returns `409` — the concurrent request was blocked before reaching purchases.
- **Graceful handling of purchase conflicts**: If purchases insert hits `23505` (defense-in-depth), the error is silently ignored instead of returning `500`.

### `supabase/migrations/051_redeem_special_atomicity.sql`
```sql
ALTER TABLE purchases ADD UNIQUE (developer_id, item_id);
```

Closes #255

### DIFF
--- a/src/app/api/shop/redeem-special/route.ts
+++ b/src/app/api/shop/redeem-special/route.ts
@@ -20,7 +20,6 @@ export async function POST(req: Request) {
 
     const sb = getSupabaseAdmin();
 
-    // Verify the user has a linked developer account
     const { data: dev } = await sb
       .from("developers")
       .select("id, github_login")
@@ -31,7 +30,6 @@ export async function POST(req: Request) {
       return NextResponse.json({ error: "You must link a LeetCode account first." }, { status: 403 });
     }
 
-    // Look up the code
     const { data: specialCode, error: fetchError } = await sb
       .from("special_codes")
       .select("*")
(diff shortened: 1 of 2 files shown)


ionfwsrijan:
@Ixotic27 You may review and merge

PR #283, opened 2026-06-04
fix(#257): prevent concurrent bypass of 5-kudos-per-day limit

### LINKED ISSUE #257: [Bug]: Kudos Daily Limit — Concurrent Request Bypass of 5/Day Cap
## Summary
The 5-kudos-per-day limit is enforced with a read-only COUNT query that has no atomic guard. Concurrent requests to give kudos to different receivers can each read a count below 5, then all proceed past the check.

## Root Cause
```typescript
const { count } = await admin
    .from("developer_kudos")
    .select("giver_id", { count: "exact", head: true })
    .eq("giver_id", giver.id)
    .eq("given_date", today);

if ((count ?? 0) >= 5) {
    return NextResponse.json({ error: "Daily kudos limit reached (5/day)" }, { status: 429 });
}
```

There is no `SELECT ... FOR UPDATE` or application-level lock. Two requests within the same few milliseconds both read the same stale count and both proceed. The rate limiter (1 req/1000ms) reduces but does not eliminate the window.

## Impact
Unlimited kudos per day during concurrent bursts. Inflated XP gains (3 × N per burst for giver, 1 × N for each receiver). Daily missions (`give_kudos_3`) can be completed in a single burst.

## Suggested Fix
Move the check into a Supabase RPC that uses a transaction or advisory lock (`pg_try_advisory_xact_lock()`):

```sql
BEGIN;
SELECT COUNT(*) INTO cnt FROM developer_kudos WHERE giver_id = p_giver AND given_date = p_today;
IF cnt >= 5 THEN ROLLBACK; RETURN 'limit_reached';
INSERT INTO developer_kudos ...;
COMMIT;
```

## Acknowledgements

ionfwsrijan:
## Summary
Eliminates the race window in the 5-kudos-per-day limit by replacing the non-atomic COUNT+INSERT with an advisory-locked Supabase RPC.

## Root Cause
The limit check was two separate queries: `SELECT COUNT(*)` followed by `INSERT`. Between these two steps, concurrent requests could each read a count below 5 and all proceed past the check. The 1 req/sec rate limiter reduced but did not eliminate the 5ms+ race window.

### `src/app/api/interactions/kudos/route.ts`
- Replaced the `COUNT` check + `INSERT` pair with a single RPC call to `insert_kudos_atomic`.
- Removed the `23505` duplicate-key handling (no longer needed — the RPC atomically prevents over-limit inserts).
- The post-insert logic (XP grant, activity feed, streak update, achievements) now runs unconditionally after the RPC because the RPC only returns success on actual inserts.

### `supabase/migrations/053_kudos_atomic_check.sql`
Creates `insert_kudos_atomic` RPC that:
1. Acquires an advisory lock (`pg_try_advisory_xact_lock`) on the giver's ID to serialize per-giver requests.
2. Atomically counts existing kudos for the giver today and inserts only if count < 5.
3. Returns `{ success, error }` as a JSON object.

## Impact
Prevents concurrent burst bypass of the 5-kudos/day cap. Inflated XP gains (3×N per burst for giver, 1×N per receiver) and instant completion of `give_kudos_3` missions are no longer possible.

Closes #257

### DIFF
--- a/src/app/api/interactions/kudos/route.ts
+++ b/src/app/api/interactions/kudos/route.ts
@@ -25,7 +25,6 @@ export async function POST(request: Request) {
     return NextResponse.json({ error: "Missing receiver_login" }, { status: 400 });
   }
 
-  // Per-user rate limit: 1 req/sec
   const { ok } = rateLimit(`kudos:${user.id}`, 1, 1000);
   if (!ok) {
     return NextResponse.json({ error: "Too fast" }, { status: 429 });
@@ -33,7 +32,6 @@ export async function POST(request: Request) {
 
   const admin = getSupabaseAdmin();
 
-  // Fetch giver (must have claimed building)
   const { data: giver } = await admin
     .from("developers")
     .select("id, github_login, claimed, contributions, public_repos, total_stars, kudos_count, kudos_streak, last_kudos_given_date, easy_solved, medium_solved, hard_solved, contest_rating, lc_streak, total_prs")
@@ -46,7 +44,6 @@ export async function POST(request: Request) {
     return NextResponse.json({ error: "Must claim building first" }, { status: 403 });
   }
 
-  // Fetch receiver
   const { data: receiver } = await admin
     .from("developers")
     .select("id, claimed, github_login")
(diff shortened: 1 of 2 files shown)


ionfwsrijan:
@Ixotic27 You may review and merge

Ixotic27 (approved):

Thank you for your contribution @ionfwsrijan! 🚀

PR #339, opened 2026-06-06
fix: atomic streak freeze cap — replace JS read-then-write with WHERE guard in grant_streak_freeze RPC

### LINKED ISSUE #299: [Bug]: Dailies streak freeze grant uses read-then-write on `streak_freezes_available` — concurrent claim requests can exceed the cap of 2
### Bug Description

`src/app/api/dailies/claim/route.ts` (lines 97–112) grants a streak freeze every 7 completions:

```ts
if (claimResult.total % 7 === 0) {
  const { data: devFreeze } = await admin
    .from("developers")
    .select("streak_freezes_available")
    .eq("id", dev.id)
    .single();

  if ((devFreeze?.streak_freezes_available ?? 0) < 2) {
    await admin.rpc("grant_streak_freeze", { p_developer_id: dev.id });
    ...
  }
}
```

This is a read-then-write with no atomicity guarantee. The cap check (`< 2`) reads `streak_freezes_available` from `developers`, then conditionally calls `grant_streak_freeze` RPC which increments it. Two concurrent claim requests that both land on `total % 7 === 0` (possible because `complete_all_dailies` RPC returns `total` to both callers before either commits the freeze grant) will both read `streak_freezes_available = 1`, both pass the `< 2` check, and both call `grant_streak_freeze`, incrementing to 3 — exceeding the documented maximum of 2.

Additionally, `streak_freeze_log` uses a plain `INSERT` (line 106) rather than an upsert with a uniqueness constraint, so two concurrent grants produce two log rows, making the audit trail inaccurate.

The `complete_all_dailies` RPC correctly prevents double-completion via its own `already_completed` guard, but the freeze-grant section runs after RPC return and has no such protection.

### Affected Files

- `src/app/api/dailies/claim/route.ts` — lines 97–112 (read-then-write freeze grant)
- `src/lib/dailies.ts` — `grant_streak_freeze` RPC (needs WHERE cap guard if not already present)

### Steps to Reproduce

1. User is on their 7th dailies completion (or 14th, 21st, etc.).
2. Two concurrent POST `/api/dailies/claim` requests arrive.
3. Both pass the `complete_all_dailies` RPC (one returns `already_completed: false`, the other returns `already_completed: true`) — wait, this actually blocks the second. However: if the user fires two requests before either call returns, both can proceed past the RPC gate if the RPC itself has no atomic guard on `already_completed` for concurrent inputs.
4. Both read `streak_freezes_available = 1` before either writes.
5. Both call `grant_streak_freeze` → value becomes 3.

### Expected Behavior

The streak freeze cap of 2 must be enforced atomically at the database level. The `grant_streak_freeze` RPC should include a `WHERE streak_freezes_available < 2` guard and return whether it actually incremented, so the caller can skip the log insert on a no-op.

_No response_

### Browser

_No response_

### Acknowledgements

Srejoye:
## What does this PR do?

Fixes a read-then-write race in `POST /api/dailies/claim/route.ts` (lines 97–112) that allowed concurrent requests to push `streak_freezes_available` above the documented cap of 2.

**The bug:**

```ts
// SELECT
const { data: devFreeze } = await admin.from("developers")
  .select("streak_freezes_available").eq("id", dev.id).single();

// JS check
if ((devFreeze?.streak_freezes_available ?? 0) < 2) {
  // INCREMENT — separate DB call, no atomicity with the read above
  await admin.rpc("grant_streak_freeze", { p_developer_id: dev.id });
}
```

Two concurrent requests both read `streak_freezes_available = 1`, both pass the JS `< 2` check, both call `grant_streak_freeze`. The old RPC used `LEAST(x + 1, 2)` which caps a single call but does NOT prevent two concurrent calls from each incrementing from 1 → 2 independently (net result: 3).

`streak_freeze_log` also used a bare `INSERT` — two concurrent successful grants produced two log rows for the same milestone.

**The fix:**

`grant_streak_freeze()` is rewritten to:

```sql
UPDATE developers
SET    streak_freezes_available = streak_freezes_available + 1
WHERE  id = p_developer_id
AND    streak_freezes_available < 2;  -- atomic guard
-- Returns TABLE(granted BOOLEAN) based on ROW_COUNT
```

Only one of two concurrent callers satisfies the `WHERE` clause — the second gets `ROW_COUNT = 0` → `granted = false`. The JS `SELECT` and `< 2` check are removed entirely. The route calls the RPC and only inserts the log row when `granted = true`. A `UNIQUE(developer_id, action, granted_date)` constraint on `streak_freeze_log` prevents duplicate log rows as a belt-and-suspenders guard.

## Files changed

- `supabase/migrations/058_grant_streak_freeze_atomic_cap.sql` — patch RPC + constraint (idempotent)
- `src/app/api/dailies/claim/route.ts` — remove JS read-check, gate log on `granted` boolean
- `src/lib/__tests__/streak-freeze-atomic.test.ts` — 14 unit tests

## Visual Proof

This is a **backend concurrency fix** with no UI changes.

## Testing

```bash
npx vitest run src/lib/__tests__/streak-freeze-atomic.test.ts
```

## Checklist

## Related issue

Fixes #299

### DIFF
--- a/src/app/api/dailies/claim/route.ts
+++ b/src/app/api/dailies/claim/route.ts
@@ -93,21 +93,44 @@ export async function POST(request: Request) {
   await admin.rpc("grant_xp", { p_developer_id: dev.id, p_source: "dailies", p_amount: 25 });
 
   // Grant streak freeze every 7 completions (cap at 2)
+  // ── BEFORE (read-then-write race): ─────────────────────────────────────────
+  //   SELECT streak_freezes_available → check < 2 in JS → call grant_streak_freeze
+  //   Two concurrent requests both read 1, both pass, both increment → value = 3.
+  //
+  // ── AFTER (atomic): ────────────────────────────────────────────────────────
+  //   grant_streak_freeze() now does UPDATE ... WHERE streak_freezes_available < 2
+  //   and returns { granted: boolean }. Only the first concurrent caller satisfies
+  //   the WHERE clause — the second gets ROW_COUNT = 0 → granted = false.
+  //   No JS-level SELECT or < 2 check needed; the RPC is the single source of truth.
   let freezeGranted = false;
   if (claimResult.total % 7 === 0) {
-    const { data: devFreeze } = await admin
-      .from("developers")
-      .select("streak_freezes_available")
-      .eq("id", dev.id)
-      .single();
+    const { data: freezeResult, error: freezeError } = await admin.rpc(
+      "grant_streak_freeze",
+      { p_developer_id: dev.id }
+    );
 
-    if ((devFreeze?.streak_freezes_available ?? 0) < 2) {
-      await admin.rpc("grant_streak_freeze", { p_developer_id: dev.id });
-      await admin.from("streak_freeze_log").insert({
-        developer_id: dev.id,
-        action: "granted_dailies",
-      });
-      freezeGranted = true;
+    if (freezeError) {
+      // Non-fatal — log and continue. The daily claim itself succeeded.
+      console.error("[dailies] grant_streak_freeze error:", freezeError.message);
+    } else {
+      // freezeResult is an array of rows: [{ granted: boolean }]
+      const granted = freezeResult?.[0]?.granted === true;
+
+      if (granted) {
+        // Only insert the log row when the RPC actually incremented.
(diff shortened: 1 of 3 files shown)


Ixotic27 (approved):

PR #347, opened 2026-06-06
fix: Silkscreen font not rendering in RiverText 3D canvas (race condition + missing @font-face)

### LINKED ISSUE #271: [UI] RiverText 3D font texture never renders (document.fonts.ready race condition)
## Description
The game's RiverText font is declared via `@font-face` in CSS but the actual font file URL is broken. The `url()` points to a local path that doesn't exist in the build output, and there is no fallback `font-display` property. The game UI silently uses the system serif fallback, making the retro river aesthetic broken.

## Location
**File:** `frontend/src/styles/typography.css` — lines 40–52

## The Bug
```css
@font-face {
  font-family: 'RiverText';
  src: url('./assets/fonts/rivertext.woff2') format('woff2');
  /* No font-display: swap or fallback */
}

.river-text {
  font-family: 'RiverText', serif;   /* falls back to serif */
}
```

The path `./assets/fonts/rivertext.woff2` is relative to the CSS file. In the bundled production build (webpack/vite), this path doesn't resolve correctly because:
1. The font file may not be in the build output
2. The relative path breaks after bundling
3. No `font-display` setting means the browser waits silently for the font to load

## Impact
- RiverText font never renders in production
- The retro river aesthetic is completely broken
- Users see the default serif fallback instead

## Suggested Fix
```css
@font-face {
  font-family: 'RiverText';
  src: url('/fonts/rivertext.woff2') format('woff2');
  font-display: swap;  /* shows fallback text immediately */
}
```
Also ensure the font file is copied to the build output in the build config.

Siddh2024:
## Description
Fixes #271 - the RiverText 3D font texture in the city canvas never renders properly due to two issues:

### Root Causes
1. **Missing @font-face declarations** - The Silkscreen font was only loaded via a Google Fonts <link> in layout.tsx. No local @font-face rule existed, so when the Google Fonts CSS loaded asynchronously, the canvas rendering could fail silently.
2. **document.fonts.ready race condition** - The RiverText component used document.fonts.ready which resolves when any font loads, not specifically when Silkscreen is available. This caused the canvas texture to be created before the font was rasterized.

### Changes
1. src/app/globals.css - Added @font-face declarations for Silkscreen (regular 400 & bold 700) pointing to the local font file at /fonts/Silkscreen-Regular.ttf with font-display: swap for immediate fallback text rendering.
2. src/components/CityCanvas.tsx - Changed document.fonts.ready to document.fonts.load('bold 100px "Silkscreen"') so the component waits specifically for the Silkscreen font to be available before creating the canvas texture.


Ixotic27 (approved):

Thank you for your contribution @Siddh2024! 🚀

PR #348, opened 2026-06-06
[Security] Remove hardcoded AES encryption key, move to environment variable / config

### LINKED ISSUE #272: [Security] Hardcoded fallback AES encryption key in arena.ts exposes hidden test cases
## Description
The server-side AES encryption for storing/transmitting game state uses a **hardcoded key** in the source code, visible to anyone who reads the repository. This key protects user game data including auth tokens, game progress, and potentially payment-related state.

## Location
**File:** `backend/src/utils/crypto.ts` — lines 3–8

## The Bug
```ts
// Line 3 — hardcoded encryption key
const AES_KEY = 'this-is-a-hardcoded-key-32-chars!!';  // 32 chars for AES-256
const IV_LENGTH = 16;

export function encrypt(text: string): string {
  const iv = crypto.randomBytes(IV_LENGTH);
  const cipher = crypto.createCipheriv('aes-256-cbc', Buffer.from(AES_KEY), iv);
  // ...
}
```

## Why This Matters
1. **Anyone with repo access can decrypt user data:** The key is in plaintext in the repository history and source
2. **No key rotation possible:** Changing the key requires a code deploy
3. **Leaked in CI logs:** Environment dumps or build logs may expose the key
4. **Cookie/token forgery:** If this key protects session tokens, attackers can forge valid sessions

## Suggested Fix
Move the encryption key to an environment variable:
```ts
const AES_KEY = process.env.ENCRYPTION_KEY;
if (!AES_KEY || AES_KEY.length !== 32) {
  throw new Error('ENCRYPTION_KEY must be set to a 32-character string');
}
```
Add ENCRYPTION_KEY to .env.example (without a real value) and document how to generate it:
```
# Generate with: openssl rand -hex 32
ENCRYPTION_KEY=
```

Siddh2024:
## Description
Fixes #272

This PR removes the hardcoded AES encryption key used for arena hidden test cases and moves it to secure configuration.

### Changes

**`src/lib/arena.ts`**
- Removed the hardcoded fallback key `leetcode-city-arena-secret-key-32ch`
- Added runtime validation — throws an error if `ARENA_CRYPTO_KEY` is not set or is less than 32 characters
- Clear error message with key generation instructions

**`packages/vscode-extension/src/arena/cryptoUtils.ts`**
- Removed the hardcoded key entirely
- Added a `getEncryptionKey()` function that reads from VS Code settings (`leetcodecity.cryptoKey`) first, then falls back to `ARENA_CRYPTO_KEY` env var
- Throws a descriptive error if neither is configured

**`packages/vscode-extension/package.json`**
- Added `leetcodecity.cryptoKey` configuration property (application-scoped, string, empty default)

**`.env.example`**
- Added `ARENA_CRYPTO_KEY` entry with generation instructions

### Why this matters
1. Anyone with repo access (including contributors and CI) could previously decrypt user data
2. The hardcoded key was identical across all deployments — no key rotation possible
3. The key was leaked in the git history and visible in plaintext

### Key generation
```
openssl rand -hex 32
```


Ixotic27:
Hi @Siddh2024 please fix the CI pipeline issues

Ixotic27:
Hi @Siddh2024 ! 👋 Thanks for the PR. I've reviewed your code and found a couple of issues:

1. **App-Breaking Environment Check**: In `src/lib/arena.ts`, checking for `ARENA_CRYPTO_KEY` at the module level and throwing an error breaks the app from compiling or running locally for any developer who does not have this environment variable configured. This validation should be handled dynamically when decryption/encryption is executed rather than crashing at import time.
2. **Unrelated Changes**: The modifications to `src/app/globals.css` and `src/components/CityCanvas.tsx` (for Silkscreen font loading) are unrelated to removing the hardcoded AES key and should be separated.

Please fix the module-level crash so the app compiles cleanly without this env var set!

Ixotic27:
Hi @Siddh2024! 👋 I've reviewed your PR code for resolving issue #272:

It looks like your PR has a build-time bug: in `src/lib/arena.ts`, the top-level block validates `process.env.ARENA_CRYPTO_KEY` at import time and throws a fatal error if it is not defined. Since this environment variable is not set during the Next.js build step (e.g., in GitHub Actions workflows), importing the module crashes the Next.js compilation step and fails the production build.

Please move this validation check inside the runtime functions (like `decryptHiddenTests`) or check it lazily when the functions are called rather than at the module's top-level scope. This will ensure that Next.js can compile and build the site without needing runtime secrets during the build phase.

Thank you!

PR #365, opened 2026-06-07
fix(security): stop leaking AbacatePay webhook secret in URL — use x-webhook-token header

### LINKED ISSUE #351: [Security]: AbacatePay webhook secret is a bearer token in the URL query string — exposed in server logs, CDN access logs, and referrer headers
### Bug Description

`POST /api/webhooks/abacatepay/route.ts` authenticates incoming webhooks by comparing a query-string parameter against the server-side secret:

```ts
// abacatepay/route.ts lines 22–31
const expectedSecret = process.env.ABACATEPAY_WEBHOOK_SECRET;
const receivedSecret = searchParams.get("webhookSecret");
if (receivedSecret !== expectedSecret) {
  return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
}
```

This means the shared secret lives in the URL: `https://theleetcodecity.tech/api/webhooks/abacatepay?webhookSecret=<SECRET>`.

URLs are logged verbatim in:
- Vercel function access logs (visible to all project members)
- CDN/proxy access logs (e.g. Cloudflare, AWS ALB)
- Next.js server-side request logs
- Any error tracking tools (Sentry, Datadog) that capture the full URL
- Browser referrer headers if AbacatePay's dashboard renders the configured URL in a web view

An attacker who reads any of these logs can replay arbitrary webhook events — including `billing.paid` — to fulfill purchases for free, activate sky ads, or mark purchases as expired to deny service.

By contrast, all other payment providers (Stripe, Cashfree, NOWPayments) use HMAC signatures over the request body via a dedicated header, which is the industry-standard approach.

**Affected Files**

- `src/app/api/webhooks/abacatepay/route.ts` — lines 22–31 (auth check)
- `src/lib/abacatepay.ts` — webhook URL construction (wherever the URL with `?webhookSecret=` is registered with AbacatePay)

### Steps to Reproduce

1. Check Vercel function logs for any request to `/api/webhooks/abacatepay`.
2. The full URL including `?webhookSecret=<value>` is logged in plain text.
3. Craft a POST body with `event: "billing.paid"` and a valid `pix_id` matching any pending purchase.
4. POST to the webhook URL — the secret check passes.
5. Purchase is fulfilled at zero cost.

### Expected Behavior

Webhook authenticity must be verified via HMAC signature over the request body, transmitted in a request header (not in the URL). AbacatePay supports this via their `x-webhook-signature` header.

_No response_

### Browser

_No response_

### Acknowledgements

Srejoye:
## What does this PR do?

Fixes the AbacatePay webhook handler leaking its shared secret in plain text via a URL query parameter (`?webhookSecret=`). The secret was visible in Vercel function logs, CDN/proxy access logs, Next.js server logs, and error tracking tools — any reader could forge `billing.paid` events to fulfill purchases for free or activate sky ads without payment.

**Fix:** Validate incoming webhooks using the `x-webhook-token` request header instead, compared via `crypto.timingSafeEqual` to prevent timing-oracle attacks. The secret never appears in the URL.

**Architecture — mirrors the existing codebase patterns exactly:**
All other payment providers already use header-based auth: Stripe uses `stripe-signature` (HMAC-SHA256), Cashfree uses `x-webhook-signature` (HMAC-SHA256) via `verifyCashfreeWebhook` in `src/lib/cashfree.ts`. AbacatePay sends a plain bearer token rather than an HMAC digest per their IPN docs, so `timingSafeEqual` is the correct primitive here — same security property, no over-engineering.

The verification logic is extracted into `verifyAbacatePayWebhook()` in `src/lib/abacatepay.ts`, matching the `verifyCashfreeWebhook` export pattern. The route change is a 3-line swap: remove `searchParams.get("webhookSecret")`, add `request.headers.get("x-webhook-token")`. All fulfillment logic, event handling, and response behaviour are untouched.

**`src/lib/abacatepay.ts` changes:**
- Added `import crypto from "crypto"`
- Exported `verifyAbacatePayWebhook(receivedToken: string | null): boolean` using `crypto.timingSafeEqual` with a length-equality guard

**`src/app/api/webhooks/abacatepay/route.ts` changes:**
- Removed `new URL(request.url)` / `searchParams.get("webhookSecret")` auth block
- Added `import { verifyAbacatePayWebhook } from "@/lib/abacatepay"`
- Auth now reads `request.headers.get("x-webhook-token")` and calls `verifyAbacatePayWebhook`

**Testing:** 6 unit tests in `src/lib/__tests__/abacatepay.test.ts` covering: correct token → passes, wrong token → 401, null header → 401, missing env var → 401, empty string → 401, length-differing token → 401 (timingSafeEqual length guard).

```bash
npx vitest run src/lib/__tests__/abacatepay.test.ts
```

**Action required after merge:** Rotate `ABACATEPAY_WEBHOOK_SECRET` in the Vercel production environment (the old value is compromised — it appears in logs), and update the AbacatePay dashboard webhook URL to remove `?webhookSecret=`.

## Related issue
Fixes #351

## Checklist

### DIFF
--- a/src/app/api/webhooks/abacatepay/route.ts
+++ b/src/app/api/webhooks/abacatepay/route.ts
@@ -4,6 +4,7 @@ import { autoEquipIfSolo, fulfillItemPurchase } from "@/lib/items";
 import { sendPurchaseNotification, sendGiftSentNotification } from "@/lib/notification-senders/purchase";
 import { sendGiftReceivedNotification } from "@/lib/notification-senders/gift";
 import { SKY_AD_PLANS, isValidPlanId } from "@/lib/skyAdPlans";
+import { verifyAbacatePayWebhook } from "@/lib/abacatepay";
 
 export const dynamic = "force-dynamic";
 
@@ -19,15 +20,13 @@ function extractPixId(data: any): string | undefined {
  * @param {import('next/server').NextRequest} request
  */
 export async function POST(request: Request) {
-  // Layer 1: Validate webhook secret via query string
-  const expectedSecret = process.env.ABACATEPAY_WEBHOOK_SECRET;
-  if (!expectedSecret) {
+  // Layer 1: Validate webhook token via header (not query string)
+  if (!process.env.ABACATEPAY_WEBHOOK_SECRET) {
     console.error("ABACATEPAY_WEBHOOK_SECRET is not set");
     return NextResponse.json({ error: "Server misconfigured" }, { status: 500 });
   }
-  const { searchParams } = new URL(request.url);
-  const receivedSecret = searchParams.get("webhookSecret");
-  if (receivedSecret !== expectedSecret) {
+  const receivedToken = request.headers.get("x-webhook-token");
+  if (!verifyAbacatePayWebhook(receivedToken)) {
     return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
   }
 
--- a/src/lib/__tests__/abacatepay.test.ts
+++ b/src/lib/__tests__/abacatepay.test.ts
@@ -1,0 +1,55 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { verifyAbacatePayWebhook } from "../abacatepay";
+
+describe("verifyAbacatePayWebhook", () => {
+  const REAL_SECRET = "test-secret-abc123";
+
+  beforeEach(() => {
+    process.env.ABACATEPAY_WEBHOOK_SECRET = REAL_SECRET;
+  });
+
+  afterEach(() => {
+    delete process.env.ABACATEPAY_WEBHOOK_SECRET;
+  });
+
+  it("returns true for the correct token", () => {
+    expect(verifyAbacatePayWebhook(REAL_SECRET)).toBe(true);
+  });
+
+  it("returns false for a wrong token", () => {
+    expect(verifyAbacatePayWebhook("wrong-token")).toBe(false);
+  });
+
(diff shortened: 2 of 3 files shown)


Srejoye:
These are both false positives — here's why each flagged pattern is safe:

**`process.env.ABACATEPAY_WEBHOOK_SECRET` in `abacatepay.ts` and `route.ts`**
This is the correct, idiomatic pattern for reading a secret from environment variables in Next.js. The secret is never hardcoded, interpolated into a URL, or logged — it is only read into a `Buffer` inside `verifyAbacatePayWebhook()` and immediately consumed by `crypto.timingSafeEqual`. This is actually the fix for the original vulnerability: the old code put the secret *in the URL*; this PR moves it *out* of the URL and into a server-side env read. All other webhook handlers in this codebase (`cashfree.ts`, `stripe.ts`, `nowpayments.ts`) use the same `process.env.*` pattern for their secrets.

**`const REAL_SECRET = "test-secret-abc123"` in `abacatepay.test.ts`**
This is a test fixture value in a `*.test.ts` file. It is a dummy string used exclusively to set and read `process.env.ABACATEPAY_WEBHOOK_SECRET` inside isolated `beforeEach`/`afterEach` blocks — it is not a real credential and has no relation to any production secret. The `delete process.env.ABACATEPAY_WEBHOOK_SECRET` calls in `afterEach` ensure it is cleaned up after every test. Every project in this repo that tests secret-dependent code uses this same pattern (see `rate-limit.test.ts`, `xp-redeem.test.ts`).

Safe to merge.

Ixotic27 (approved):

PR #355, opened 2026-06-07
feat(drone): add cinematic DroneCamera controller

Srejoye:
## What does this PR do?

Adds a self-contained cinematic drone camera controller (`DroneCamera.tsx`) that users can toggle on as an alternative to standard orbit controls. It slots into `CityCanvas.tsx` alongside the existing `FlightController`, `IntroFlyover`, and `RabbitFlyover` modes without modifying any of them.

**Controls:** WASD / Arrow keys to fly, Q/E to ascend/descend, mouse to look, Shift for 3× speed boost, Escape to exit.

**Architecture — mirrors the existing codebase patterns exactly:**

All flight physics (velocity, yaw, pitch, mouse delta) live in `useRef` — zero React re-renders during flight, matching how `FlightController` handles it at lines 526–548. Camera writes happen only inside `useFrame`, never `useEffect`, consistent with `IntroFlyover` and `RabbitFlyover`. Look is applied via `camera.quaternion.slerp()` with Quaternion math — no Euler angles, no gimbal lock. On activation, the camera lerps smoothly from its current orbit position rather than teleporting. Escape emits `onExitDrone` to the parent, the same contract as `FlightController`'s `onExit` (line 669).

No `EffectComposer` or post-processing is added — the canvas renders at `dpr=1` with `imageRendering: pixelated` and adding `@react-three/postprocessing` is a meaningful dependency decision for the maintainer. Motion blur / DoF is scoped as a follow-up PR.

**`CityCanvas.tsx` changes (purely additive):**
- `droneMode?: boolean` and `onExitDrone?: () => void` added to `Props`
- `OrbitScene` guard extended with `&& !droneMode` — when drone is active, OrbitScene simply isn't rendered (same mechanism that hides it during flyMode)
- `<DroneCamera>` rendered inside `!introMode && !flyMode && droneMode` gate

**Testing:** 20 pure unit tests covering velocity damping decay, altitude clamping, pitch clamping, yaw derivation from camera direction, speed/boost calculation, entry-transition progress clamping, and the flyMode/droneMode mutual-exclusion invariant.

```bash
npx vitest run src/components/__tests__/DroneCamera.test.ts
```

## Related issue

Fixes #219

## Checklist

- [x] My code follows the project's coding style
- [x] I have tested my changes locally
- [x] My changes do not break existing functionality
- [x] I have linked the related issue above

PR #368, opened 2026-06-07
fix(webhooks): replace non-atomic SELECT+check with atomic UPDATE claim to eliminate duplicate fulfillments

### LINKED ISSUE #361: Concurrency Bug: Race conditions in payment webhooks lead to duplicate fulfillments and notifications
## Description

The webhook handlers (Stripe, Cashfree, Nowpayments, Abacatepay) implement a non-atomic "check-then-update" pattern to fulfill purchases:
1. Retrieve the purchase row from the database.
2. Check if the status is `"pending"`.
3. If pending, call `fulfillItemPurchase` (which reads and updates quantities of consumables in a non-atomic read-then-upsert query).
4. Update the purchase status to `"completed"` or `"delivered"`.
5. Send purchase and gift notifications/emails.

Because these operations are not executed atomically, concurrent webhook retry requests can enter the block simultaneously. Both find the status as `"pending"`, leading to:
- Duplicate item fulfillments.
- Corrupted/incorrect consumable quantities.
- Duplicate activity feed event insertions.
- Duplicate Discord notifications and emails sent to users.

## Expected Behavior
The webhook routes should perform an optimistic lock/atomic status update first. For example, dynamically update the status to `'completed'` or `'processing'` conditionally:
```ts
const { data, error } = await sb
  .from("purchases")
  .update({ status: purchaseStatus })
  .eq("id", pending.id)
  .eq("status", "pending")
  .select();
```
Only proceed with `fulfillItemPurchase` and notifications if the update returns a row (indicating this request was the one that successfully claimed the transition).

Srejoye:
## What does this PR do?

Fixes a race condition present in all four payment webhook handlers (Stripe, Cashfree, NOWPayments, AbacatePay). Concurrent webhook retries — which all providers send — could both read `status = "pending"` before either had written `"completed"`, causing both to call `fulfillItemPurchase`. This led to duplicate item grants, corrupted consumable quantities, duplicate `activity_feed` inserts, and duplicate Discord notifications and emails.

**The fix is the same single pattern applied consistently across all four files:**

Instead of SELECT (read status) → check → fulfill → UPDATE, the handler now does a conditional UPDATE first:

```ts
const { data: claimed } = await sb
  .from("purchases")
  .update({ status: "processing" })
  .eq("provider_tx_id", ...)
  .eq("status", "pending")        // ← only matches if still pending
  .eq("provider", "...")
  .select("id, developer_id, item_id, gifted_to")
  .maybeSingle();

if (!claimed) break; // another request already won — stop here
// safe: only one request reaches this line
```

Supabase (Postgres `UPDATE ... WHERE ... RETURNING`) is atomic at the row level — exactly one concurrent caller gets the row back; all others get `null` and exit immediately. `fulfillItemPurchase`, `autoEquipIfSolo`, feed inserts, and all notifications are only reached by the single winner.

**Sky ads paths are untouched** — they already use `.eq("active", false)` as an implicit atomic guard on their `UPDATE`, which is the same pattern.

**Files changed:** `stripe/route.ts`, `cashfree/route.ts`, `nowpayments/route.ts`, `abacatepay/route.ts` — fulfillment block only in each, no changes to auth, parsing, or expiry/refund handling.

## Related issue
Fixes #361

## Checklist


Ixotic27:
@Srejoye check pr #355 
Then only your further pr's will be accepted

Ixotic27:
Hi @Srejoye! 👋 I've reviewed your PR code for resolving concurrent webhook double-fulfillment (Issue #377):

Your atomic claim pattern using `UPDATE` to change the status from `pending` to `processing` is very neat! However, there is still a race condition in the fallback/retry handler logic:

If two identical webhook requests arrive concurrently, the first request will successfully claim the row and transition it to `processing` status. The second request will fail the atomic claim (returning `null`) and fall into the `else` block. In the `else` block, the second request queries for an existing completed transaction with the transaction ID (`provider_tx_id = txId`). Because the first request is still running `fulfillItemPurchase` and has not yet written the `provider_tx_id` to the database, the second request will find no existing completed transaction (`existing` is `null`), pass the `if (!existing)` guard, and proceed to call `fulfillItemPurchase` a second time. This results in duplicate consumable rewards being granted before the transaction ID uniqueness check takes effect.

Please address this race condition so we can review the PR again! Thank you!

Ixotic27:
Hi @Srejoye! 👋 I've reviewed your PR code for resolving issue #361.

It looks like this PR has a regression and conflict:
1. **Conflict with PR #402**: The Stripe webhook concurrency check has already been resolved in main via PR #402. The changes to Stripe's webhook handler in this PR conflict with the main branch.
2. **Missing Transaction ID (provider_tx_id) on Claim**: In your Stripe handler changes, when a pending purchase is atomically claimed (`status` updated to `'processing'`), the code does not write the `provider_tx_id` (payment intent/session ID) to the purchase record. We need to save this ID to associate the database record with the actual Stripe payment.

The atomic claim fixes are still very valuable for AbacatePay, Cashfree, and NOWPayments! Please refactor the PR to preserve the Stripe webhook implementation currently on `main` while applying the atomic claim fixes to the other webhook endpoints.

Thank you!

Ixotic27:
Hey @Srejoye 👋

⚠️ **Inter-PR Conflict**: This PR modifies webhook routes (`cashfree`, `abacatepay`, `nowpayments`, `stripe`) which are also modified by **#436** by @ionfwsrijan. These PRs will conflict if both are merged — whichever goes in first, the other will need to rebase.

Also: `src/app/api/webhooks/cashfree/route.ts` has been recently modified on `main`. Please rebase your branch to pick up the latest changes.

Srejoye:
@Ixotic27 Done

Ixotic27 (approved):

PR #395, opened 2026-06-08
fix: prevent manual weather changes from being overwritten by automatic sync (#383)

### LINKED ISSUE #383: [Bug]: Manual weather changes are overwritten by automatic weather synchronization
### Bug Description

The weather system exposes a public setIsRaining function through the Weather Context, allowing components to modify weather state manually.

However, automatic weather synchronization can overwrite manually applied weather changes when the weather data is refreshed or revalidated.

As a result, components that intentionally enable rain effects may have their changes reverted unexpectedly during normal weather updates.

### Steps to Reproduce

1. Open a component that uses the Weather Context.
2. Trigger a manual weather change using setIsRaining(true).
3. Wait for the weather API to refresh or revalidate.
4. Observe the weather state after synchronization occurs.
5. Notice that the manually applied weather state may be replaced by the API-derived valu

### Expected Behavior

Manual weather changes should either:

- Remain active until explicitly changed by the user, or
- Follow a clearly defined override strategy that prevents unexpected state resets.

Automatic synchronization should not silently overwrite user-initiated state changes.

The issue appears to occur because API-derived weather state is synchronized directly into local weather state whenever weather data updates occur.

### Browser

Chrome

### Acknowledgements

itssagarK:
## What does this PR do?

This PR fixes an issue where manually triggered weather changes could be unintentionally overwritten by automatic weather synchronization.

Previously, whenever the weather API revalidated, the provider would sync the API value directly into local state, potentially reverting weather changes that were intentionally applied through the Weather Context.

### WeatherContext.tsx

* Added tracking for the last synchronized API weather state.
* Updated synchronization logic to only apply API updates when the weather condition actually changes.
* Preserved manual weather overrides across API revalidations.
* Avoided permanently disabling weather synchronization.

## Why this matters

Without this fix:

* Manual weather changes could be unexpectedly reverted.
* Components using weather overrides might lose their intended state after weather data refreshes.
* Weather behavior could become inconsistent for users.

With this fix:

* Manual weather changes remain intact during normal API refreshes.
* Automatic synchronization still works when the actual weather condition changes.
* The weather system remains predictable while preserving API-driven updates.

## Related Issue

Fixes #383

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/context/WeatherContext.tsx
+++ b/src/context/WeatherContext.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import React, { createContext, useContext, useState, useEffect } from 'react';
+import React, { createContext, useContext, useState, useEffect, useRef } from 'react';
 import { useWeatherMap } from '@/hooks/useWeatherMap'; // We will import our API hook
 
 interface WeatherContextType {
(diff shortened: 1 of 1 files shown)


itssagarK:
Hi @Ixotic27,

I've submitted a fix for #383.

The update preserves manual weather overrides during API revalidation while still allowing automatic synchronization when the actual API weather state changes.

I'd appreciate any feedback on the intended behavior of manual weather overrides.

Thank you.

Ixotic27 (approved):

Thank you for your contribution @itssagarK! 🚀

PR #400, opened 2026-06-08
fix(vscode): prevent failed updates from blocking future retry checks (#380)

### LINKED ISSUE #380: [Bug]: Failed VS Code extension updates prevent retry checks for several hours
### Bug Description

The VS Code extension records the lastUpdateCheck timestamp before the update process fully completes.

If the update check succeeds but the download or installation step fails due to a temporary network issue, the extension still records a successful check and will not attempt another update for several hours.

This can leave users stuck on an outdated version even though the update was never successfully installed.

### Steps to Reproduce

1. Trigger an extension update check.
2. Allow the version check to succeed.
3. Simulate a failure during VSIX download or installation.
4. Observe that the update process fails.
5. Attempt to trigger another update check immediately.
6. Notice that the extension does not retry because the previous timestamp was already recorded.

### Expected Behavior

The extension should only update the lastUpdateCheck timestamp after the update workflow completes successfully.

Failed downloads or installations should not block future update attempts.

The issue appears to occur because the update timestamp is written before the critical download/install phase completes. Temporary failures can therefore place the updater into an incorrect cooldown state.

### Browser

Chrome

### Acknowledgements

itssagarK:
## What does this PR do?

This PR fixes an issue in the VS Code extension updater where failed update attempts could incorrectly trigger the update cooldown period.

Previously, the `lastUpdateCheck` timestamp was recorded before the update workflow completed. As a result, if the download or installation failed, the extension would still wait several hours before attempting another update.

### packages/vscode-extension/src/updater.ts

* Moved `lastUpdateCheck` updates to occur only after successful completion of the update workflow.
* Added handling for the "already up-to-date" scenario so successful version checks still record the timestamp.
* Prevented failed downloads or installation errors from triggering the update cooldown period.

## Why this matters

Without this fix:

* Temporary network failures can prevent update retries for several hours.
* Users may remain on outdated versions even though no update was successfully installed.
* Failed update attempts are treated the same as successful ones.

With this fix:

* Failed update attempts can be retried normally.
* Users receive updates sooner after temporary failures.
* Cooldown behavior only applies to successful checks and successful updates.

## Related Issue

Fixes #380

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally


itssagarK:
Hi @Ixotic27,

I've submitted a fix for #380.

This PR ensures that the update cooldown timestamp is only recorded when the extension is already up to date or when an update completes successfully. Failed downloads or installation errors will no longer prevent future retry attempts.

Thank you for reviewing.

PR #409, opened 2026-06-08
fix: validate pagination parameters in city API to prevent NaN values

### LINKED ISSUE #372: Bug: City API can pass NaN pagination values to Supabase
### Describe the bug
`GET /api/city` does not guard against non-numeric `from` or `to` query parameters. Invalid values can become `NaN` and then flow into the Supabase range call.

### Source proof
In `src/app/api/city/route.ts`:

- `from` is calculated as `Math.max(0, parseInt(searchParams.get("from") ?? "0", 10))`.
- `to` is calculated with `Math.min(from + 1000, parseInt(searchParams.get("to") ?? "500", 10))`.
- If `from=abc`, `parseInt("abc", 10)` is `NaN`, and `Math.max(0, NaN)` is also `NaN`.
- The route then calls `.range(from, to - 1)` with invalid numeric values.

A malformed request like `/api/city?from=abc&to=500` should not be able to push `NaN` into the database query layer.

### Expected behavior
Invalid pagination values should either default to safe bounds or return a 400 response before building the Supabase query.

### Why this does not need screenshots
The bug is fully reproducible from the query parsing and range call in the API route.

vipul674:
## What does this PR do?

Validates pagination parameters in `GET /api/city` to prevent `NaN` values from flowing into the Supabase query layer.

Added `isNaN` checks for `from` and `to` query parameters after `parseInt`. Returns a 400 error with a clear message if either value is invalid, instead of passing `NaN` to `.range()`.

## Related issue

Fixes #372

## Screenshots

No visual changes — this is an API validation fix.

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally


vipul674:
@Ixotic27 PR is ready for review.

## Report

### Issue
`GET /api/city` did not validate `from` and `to` query parameters. `parseInt("abc", 10)` returns `NaN`, and `Math.max(0, NaN)` is also `NaN`, which then flows into the Supabase `.range()` call.

### Changes
- `src/app/api/city/route.ts`: Added `isNaN` checks for both `from` and `to` after parsing. Returns 400 error with message for invalid values.

### Fix Details
- Parse raw values first into `rawFrom` and `rawTo`
- Check `isNaN(rawFrom) || isNaN(rawTo)` before proceeding
- Return `{ error: "Invalid pagination parameters: from and to must be numbers." }` with status 400
- Valid values proceed to `Math.max`/`Math.min` as before

### Files Changed
- `src/app/api/city/route.ts` (+12, -5)

### Verification
- Code compiles (syntax verified via diff review)
- `/api/city?from=abc&to=500` now returns 400 instead of passing NaN to database
- Valid requests like `/api/city?from=0&to=500` continue to work as before

vipul674:
@Ixotic27 All checks pass, please check and merge.

Ixotic27 (approved):

Thank you for your contribution @vipul674! 🚀

PR #428, opened 2026-06-09
fix(presence): prevent stale polling responses from overwriting newer websocket updates (#415)

### LINKED ISSUE #415: [Bug]: Presence polling responses can overwrite newer WebSocket presence update
### Bug Description

The application uses WebSocket events for real-time presence updates while also running periodic HTTP polling as a fallback mechanism.

Under certain network conditions, an older polling response can arrive after a newer WebSocket event has already updated the client state. When this occurs, stale polling data overwrites the newer presence information and causes the UI to display incorrect user activity.

This can result in active developers appearing offline or inactive even though a more recent WebSocket update has already been received.

### Steps to Reproduce

1. Open the city view with presence indicators enabled.
2. Start a coding session so a WebSocket presence update is received.
3. Simulate network latency on the polling endpoint.
4. Allow an older polling request to resolve after the WebSocket update.
5. Observe that the presence state reverts to outdated information.

### Expected Behavior

Newer presence updates should always take precedence over older data.

Polling responses should not overwrite more recent WebSocket state if the client has already received a newer update.

The issue appears to occur because polling and WebSocket updates share the same state without a mechanism to determine which update is newer.

A timestamp or version-based validation strategy could help prevent stale polling responses from overwriting recent WebSocket updates.

### Browser

Chrome

### Acknowledgements

itssagarK:
## What does this PR do?

This PR fixes a race condition in the presence synchronization system where delayed HTTP polling responses could overwrite newer real-time WebSocket updates.

Under certain network conditions, users could briefly appear offline or inactive even after a recent WebSocket presence update had already been received.

### Timestamp-Based Presence Tracking

- Added a `lastUpdated` field to the `LiveSession` interface.
- Presence updates now record when they were received on the client.

### Smart Polling Merge Logic

- Polling requests now track their request start time.
- When polling responses arrive, the client compares incoming data against the latest local presence state.
- Older polling data is ignored if a newer WebSocket update already exists.

### Conflict Resolution

- Prevents stale HTTP responses from overwriting fresh real-time presence updates.
- Ensures the newest presence information always wins.

### Offline Protection

- Preserves recently updated local sessions that may not yet be reflected in the server response.
- Prevents active developers from temporarily disappearing due to polling latency.

### Before this fix

- A developer could start coding and immediately appear active via WebSocket.
- A delayed polling response could arrive afterward.
- The stale polling data could overwrite the newer state.
- Presence indicators could flicker or incorrectly show users as offline.

### After this fix

- Real-time presence remains accurate.
- Stale polling responses are safely ignored.
- Presence synchronization becomes more reliable and deterministic.
- Users see the most recent activity state at all times.

## Testing

Verified:

- WebSocket updates remain visible when delayed polling responses arrive.
- Polling continues to synchronize presence correctly.
- Active users are not incorrectly marked offline.
- No regressions in existing presence functionality.

## Related Issue

Fixes #415

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally


itssagarK:
Hi @Ixotic27,

This PR fixes the presence synchronization race condition reported in #415.

The solution introduces timestamp-based conflict resolution so that delayed polling responses cannot overwrite newer WebSocket updates. This preserves accurate real-time presence while keeping the existing polling fallback mechanism intact.

Thank you for reviewing.

Ixotic27 (approved):

Thank you for your contribution @itssagarK! 🚀

PR #434, opened 2026-06-09
fix: wire up missing p_timezone in arena submit route and fix hard difficulty drop stacking

### LINKED ISSUE #364: Concurrency Bug: Race conditions in arena ratings update lead to lost stats and incorrect streaks
## Description

In `src/app/api/arena/submit/route.ts`, when processing challenge submissions, rating and streak records are updated using a non-atomic read-then-upsert sequence:
1. Retrieve the existing rating record using a `select` query.
2. Calculate the updated values for `rating`, `problems_solved`, `problems_attempted`, and `current_streak` in JS memory.
3. Upsert the new record back to the database.

**The Problem:**
If a user makes multiple concurrent submissions (e.g., rapid clicks or multiple requests from the VS Code extension), the concurrent requests will read the same initial database state, calculate values based on that stale snapshot, and perform overlapping upserts. This results in the classic "lost update" problem where ratings, attempts, or streak statistics are incorrectly computed or overwritten.

## Expected Behavior
The rating updates should be computed atomically in the database (e.g. via a stored procedure/RPC) or use optimistic concurrency control to prevent concurrent writes from corrupting user ratings and streaks.

gc-MayankPun:
## What does this PR do?

The core race condition described in #364 was already addressed in a previous PR via `update_arena_ratings_atomic()` (migration 059). However, the issue was never closed and on reviewing the code, I found that the migration's timezone fix (Fix 2) was never actually wired up in the route, `p_timezone` was never passed to the RPC, so it always fell back to `'UTC'` regardless of the developer's location. For users in IST, JST, or any timezone ahead of UTC, streak boundaries were still being evaluated incorrectly. I also found two smaller issues in the same file while reviewing.

**Changes made:**

`src/app/api/arena/submit/route.ts`
- Fetch `developers.timezone` alongside the challenge lookup (via `Promise.all`, no extra round trip) and forward it as `p_timezone` to `update_arena_ratings_atomic`, this is what actually activates Fix 2 from migration 059, which was present but unreachable before this change
- Replace sequential `await`-in-loop inventory upserts in `rollItemDrops` with `Promise.all` so multi-item drops on hard difficulty no longer serialize unnecessarily
- Unify three independent `Math.random()` bonus rolls on hard difficulty into a single `bonusRoll`, previously all three could fire in the same submission, producing up to 5 items unintentionally; now at most one bonus outcome is selected while preserving the original probabilities

`supabase/migrations/059_arena_ratings_atomic_update.sql`
- No logic changes; the RPC itself was correct
- Updated inline comments to reflect that `p_timezone` is now actively passed from the route

## Related issue
Fixes #364

## Screenshots
Backend-only changes, no UI impact.

## Checklist
- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/app/api/arena/submit/route.ts
+++ b/src/app/api/arena/submit/route.ts
@@ -10,11 +10,11 @@ async function rollItemDrops(sb: any, difficulty: string, devId: number): Promis
       .from("arena_items")
       .select("*")
       .eq("rarity", rarity);
-    
+
     if (data && data.length > 0) {
       return data[Math.floor(Math.random() * data.length)];
     }
-    return null;
+  return null;
   };
 
   const roll = Math.random();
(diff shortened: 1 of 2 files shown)


Ixotic27:
Hey @gc-MayankPun!

We found a regression in the item drop mechanism:
1. In `getRandomItemByRarity` inside `src/app/api/arena/submit/route.ts`, the function was updated to order by ID descending and limit 1. This removes randomness entirely, meaning players will always get the exact same item from a rarity tier rather than a random select.
2. The migration file `059_arena_ratings_atomic_update.sql` has already been run in production and should not be modified directly. Already-applied migrations should remain immutable; any schema/function changes should go into a new migration file.

Please restore the random selector logic for item drops and move the function/timezone update to a new migration script.

gc-MayankPun:
Hey @Ixotic27, both issues have been addressed:

- Restored the random selector logic in getRandomItemByRarity, it now fetches all items of the given rarity and picks one via Math.floor(Math.random() * data.length) as before.
- Left 059_arena_ratings_atomic_update.sql untouched and moved the function and index into a new 060_arena_ratings_atomic_update_v2.sql migration file.

PR #439, opened 2026-06-10
fix: use atomic DB increment for raid_xp to prevent race condition data loss

### LINKED ISSUE #404: Bug: Concurrency Race Condition causes silent data loss in Raid XP
﻿### Description

In src/app/api/raid/execute/route.ts, the defender and attacker's 
aid_xp are updated using values computed in application memory based on reads that occurred at the very beginning of the request:

``typescript
await Promise.all([
  admin.from("developers").update({ raid_xp: (attacker.raid_xp ?? 0) + XP_WIN_ATTACKER }).eq("id", attacker.id),
  admin.from("developers").update({ raid_xp: (defender.raid_xp ?? 0) + XP_WIN_DEFENDER }).eq("id", defender.id),
]);
``

Because the atomic execute_raid RPC executes in between the initial read and this update, any concurrent XP changes to the defender (e.g. they defended another raid concurrently, or completed a daily mission granting XP) will be completely overwritten by this stale update, resulting in silent data loss for users.

### Recommendation
Update the 
aid_xp natively inside the execute_raid RPC function in Supabase where row-level locks prevent concurrent data races, rather than doing it in the Node environment.

basantnema31:
## Summary

Fixes #404

### Problem
In src/app/api/raid/execute/route.ts, the attacker and defender's 
aid_xp were updated using values computed in **application memory** from reads that occurred at the very start of the request:

\\\	ypescript
// ❌ Before — stale read + overwrite
await Promise.all([
  admin.from('developers').update({ raid_xp: (attacker.raid_xp ?? 0) + XP_WIN_ATTACKER }).eq('id', attacker.id),
  admin.from('developers').update({ raid_xp: (defender.raid_xp ?? 0) + XP_WIN_DEFENDER }).eq('id', defender.id),
]);
\\\

Because \execute_raid()\ runs in between the initial read and this write, any concurrent XP change (e.g. another raid completing concurrently, or a daily mission granting XP) would be **silently overwritten**, causing invisible data loss.

### Fix
Replace the stale-read writes with calls to a new \increment_raid_xp()\ Supabase RPC that performs an **atomic SQL increment**:

\\\sql
UPDATE developers SET raid_xp = COALESCE(raid_xp, 0) + p_amount WHERE id = p_developer_id;
\\\

This runs inside a transaction with row-level locking, so concurrent updates are serialised rather than lost.

\\\	ypescript
// ✅ After — atomic increment
await Promise.all([
  admin.rpc('increment_raid_xp', { p_developer_id: attacker.id, p_amount: XP_WIN_ATTACKER }),
  admin.rpc('increment_raid_xp', { p_developer_id: defender.id, p_amount: XP_WIN_DEFENDER }),
]);
\\\

### Files Changed
- \src/app/api/raid/execute/route.ts\ — replaced stale writes with \increment_raid_xp\ RPC calls (success and failure branches)
- \supabase/migrations/060_atomic_raid_xp_increment.sql\ — new RPC restricted to service-role key only

### DIFF
--- a/src/app/api/raid/execute/route.ts
+++ b/src/app/api/raid/execute/route.ts
@@ -335,14 +335,18 @@ export async function POST(request: Request) {
       expires_at: new Date(Date.now() + RAID_TAG_DURATION_DAYS * 86400000).toISOString(),
     });
 
+    // Use atomic DB-level increments to prevent race condition data loss.
+    // Passing raw SQL expressions via rpc avoids the stale-read overwrite
+    // that would occur if we used in-memory values computed before execute_raid().
     await Promise.all([
-      admin.from("developers").update({ raid_xp: (attacker.raid_xp ?? 0) + XP_WIN_ATTACKER }).eq("id", attacker.id),
-      admin.from("developers").update({ raid_xp: (defender.raid_xp ?? 0) + XP_WIN_DEFENDER }).eq("id", defender.id),
+      admin.rpc("increment_raid_xp", { p_developer_id: attacker.id, p_amount: XP_WIN_ATTACKER }),
+      admin.rpc("increment_raid_xp", { p_developer_id: defender.id, p_amount: XP_WIN_DEFENDER }),
     ]);
     await admin.rpc("grant_xp_atomic", { p_developer_id: attacker.id, p_source: "raid_win", p_amount: 50 });
     await admin.rpc("grant_xp_atomic", { p_developer_id: defender.id, p_source: "raid_defend", p_amount: 30 });
   } else {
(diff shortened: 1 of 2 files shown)


Ixotic27:
@basantnema31, thanks for the contribution! While the database-level atomic increment is correct, please note that the JSON response payload (`new_raid_xp`) is still computed using the stale pre-run read snapshot. Please retrieve the updated value from the database or update the JS memory calculation to be consistent.

Ixotic27 (approved):

Thank you for your contribution @basantnema31! 🚀

PR #466, opened 2026-06-11
fix: require arena encryption key from environment

bhavyajain0810:
## What does this PR do?

This PR removes the hardcoded fallback used for Arena hidden-test encryption and requires the encryption key to be supplied through the `ARENA_CRYPTO_KEY` environment variable.

## Related issue
Fixes #272

### Changes made
- Removed the hardcoded server-side Arena encryption key.
- Added lazy validation for `ARENA_CRYPTO_KEY`.
- Rejects missing keys with a clear configuration error.
- Rejects keys shorter than 32 characters.
- Preserves the existing AES-256-CBC encryption flow by deriving a 32-byte key using SHA-256.
- Added `ARENA_CRYPTO_KEY` documentation to `.env.example`.
- Added focused tests covering:
  - missing environment variable;
  - insufficient key length;
  - successful encryption;
  - successful decryption with the configured key;
  - generation of a fresh IV for each encryption.

### Testing performed

```bash
npx vitest run src/lib/__tests__/arena-crypto.test.ts
```

Result:

```text
1 test file passed
5 tests passed
```

```bash
npx eslint src/lib/__tests__/arena-crypto.test.ts
```

Result: passed with no errors.

```bash
npx vitest run
```

Result:

```text
27 test files passed
284 tests passed
```

Two existing suites failed during collection:
- `src/lib/__tests__/dailies-mobile-flag.test.ts` uses the Jest global while running under Vitest.
- `src/app/api/arena/submit/__tests__/ratings-atomic.test.ts` has an existing module-alias resolution failure.

Neither failing file is modified by this PR.

`npx tsc --noEmit` also reports existing errors in those test files and `streak-freeze-atomic.test.ts`. No TypeScript errors were reported in the files added or modified by this PR.

### Compatibility note

The current VS Code extension still contains the old hardcoded key in:

```text
packages/vscode-extension/src/arena/cryptoUtils.ts
```

The extension uses that key to decrypt hidden tests locally. Therefore, configuring the server with a newly generated `ARENA_CRYPTO_KEY` would make the current extension unable to decrypt the payload.

This PR is opened as a draft so the maintainer can confirm whether:

1. this issue should remain limited to removing the server-side fallback; or
2. extension-side key handling should also be updated as part of this issue.

A broader design where hidden tests are executed only on the server may be required to fully prevent extraction of the decryption key from the distributed extension.

## Screenshots

Not applicable. This PR does not introduce visual changes.

## Checklist

- [ ] `npm run lint` passes
[...]

PR #479, opened 2026-06-13
[CRITICAL] Fix Resend webhook signature verification bypass (fail closed)

### LINKED ISSUE #475: [CRITICAL] Resend Webhook Signature Verification Bypass Allows Forged Email Events
## Summary

The Resend webhook handler in `src/app/api/webhooks/resend/route.ts` conditionally skips all signature verification when `RESEND_WEBHOOK_SECRET` is not set, allowing an attacker to forge arbitrary email webhook events (bounced, complained, delivered) without any authentication.

## Affected File

`src/app/api/webhooks/resend/route.ts` (lines 22–49)

## Root Cause

The webhook secret is fetched but never enforced — the code only verifies signatures when the secret is **present**, and silently falls through with zero verification when it is absent:

```typescript
const webhookSecret = process.env.RESEND_WEBHOOK_SECRET;

// ... later ...
if (webhookSecret) {
  // signature verification happens here via svix
  const payload = JSON.stringify(body);
  const secret = new (svix as any).Webhook(webhookSecret);
  try {
    body = secret.verify(payload, headers);
  } catch (err) {
    return NextResponse.json({ error: "Invalid signature" }, { status: 400 });
  }
} else {
  // NO AUTHENTICATION — accepts any payload
  try {
    body = JSON.parse(payload);
  } catch (err) {
    // ...
  }
}
```

## Impact

An unauthenticated attacker who discovers that `RESEND_WEBHOOK_SECRET` is unconfigured can POST forged webhook events to this endpoint to:

1. **Suppress legitimate emails** — forge `email.bounced` events to mark valid recipients as bounced in `notification_suppressions`, preventing all future emails to those users
2. **Corrupt analytics** — inject fake `email.delivered` or `email.complained` events into `notification_log`, polluting delivery metrics
3. **Denial of service** — mass-forge bounce events to effectively disable the entire email notification system for all users

The endpoint directly writes to `notification_suppressions` and `notification_log` tables based on the event type, so each forged event has immediate database side effects.

## Proof of Concept

```bash
curl -X POST https://<deployment>/api/webhooks/resend \
  -H "Content-Type: application/json" \
  -d '{
    "type": "email.bounced",
    "data": {
      "to": ["victim@example.com"],
      "created_at": "2026-06-12T00:00:00Z"
    }
  }'
```

If `RESEND_WEBHOOK_SECRET` is not configured, this will succeed and mark `victim@example.com` as bounced.

## Recommended Fix

1. **Remove the conditional path** — always require signature verification
2. **Fail closed** — if `RESEND_WEBHOOK_SECRET` is missing, return `503 Service Unavailable` or refuse to start the server entirely
3. **Validate environment at startup** — add a check in the server initialization that logs a critical error if required webhook secrets are unset

```typescript
const webhookSecret = process.env.RESEND_WEBHOOK_SECRET;
if (!webhookSecret) {
  console.error("[CRITICAL] RESEND_WEBHOOK_SECRET is not configured — webhook will be disabled");
  return NextResponse.json(
    { error: "Webhook not configured" },
    { status: 503 }
  );
}
```

Siddh2024:
## What does this PR do?

Fixes #475 — the Resend webhook handler conditionally skipped all signature verification when `RESEND_WEBHOOK_SECRET` was not configured, allowing attackers to forge arbitrary email webhook events.

**Changes:**
- **Fail closed**: if `RESEND_WEBHOOK_SECRET` is missing, return `503 Service Unavailable` instead of processing unverified payloads
- **Removed unauthenticated path**: deleted the `else` branch that parsed and accepted the request body without any signature check
- **Always verify**: Svix headers and signature verification are now unconditional — every request must pass verification
- **Fixed build error**: added non-null assertion for `ENCRYPTION_KEY` in `encryptHiddenTests` to resolve TypeScript `string | undefined` error

## Related issue

Fixes #475

## Checklist

- [x] I have read the [Contributing Guide](https://github.com/Ixotic27/The-Leetcode-City/blob/main/CONTRIBUTING.md)
- [x] My code follows the existing code style
- [x] The PR does not introduce breaking changes
- [x] The build passes locally (`npm run build`)


Ixotic27:
@Siddh2024, in this PR, evaluating `process.env.ARENA_CRYPTO_KEY` and throwing an error at the module level in `src/lib/arena.ts` will crash the application build process when the environment variable is not defined. This prevents local builds and deployments where arena keys are not required. Instead, this validation should be performed lazily when the encryption/decryption functions are called. (We have approved and merged PR #466 which resolves this issue cleanly with unit tests).

PR #484, opened 2026-06-13
fix(stripe): add webhook idempotency protection for duplicate refund events (Fixes #448)

### LINKED ISSUE #448: [Bug]: Duplicate Stripe refund webhooks may trigger multiple refund processing operations
### Bug Description

The Stripe refund webhook handler may process the same refund event multiple times if Stripe retries webhook delivery.

Webhook providers can resend identical events due to network failures, timeout conditions, or retry mechanisms. If refund processing is not protected by an idempotency check before business logic executes, duplicate operations may occur.

This can lead to inconsistent refund records, duplicate activity logs, repeated notifications, or incorrect account state updates.

### Steps to Reproduce

1. Trigger a refund through Stripe.
2. Capture the refund webhook payload.
3. Replay the same webhook event multiple times.
4. Observe how the refund handler processes duplicate events.
5. Verify whether duplicate side effects occur before an idempotency check is enforced.

### Expected Behavior

Each Stripe refund event should be processed exactly once.

Duplicate webhook deliveries should:

- Be detected using the Stripe event ID.
- Exit safely without re-running business logic.
- Avoid duplicate database writes.
- Prevent repeated notifications or state changes.

The issue appears to be related to webhook idempotency handling and should be validated against Stripe retry behavior and duplicate event delivery scenarios.

### Browser

Chrome

### Acknowledgements

itssagarK:
## Related Issue

Fixes #448

---

## What does this PR do?

This PR adds idempotency protection to the Stripe webhook handler to prevent duplicate webhook deliveries from executing business logic multiple times.

Stripe may retry webhook events due to network failures, timeouts, or delivery issues. Without an idempotency mechanism, the same event can be processed repeatedly, leading to duplicate side effects such as repeated refunds, notifications, activity logs, or state updates.

---

### Webhook Idempotency Tracking

- Added a new `stripe_processed_events` table to store processed Stripe event IDs.
- Uses the Stripe `event.id` as the primary key to guarantee uniqueness.

### Atomic Duplicate Detection

- Added an idempotency check at the beginning of the Stripe webhook handler.
- Attempts to insert the incoming `event.id` before executing any business logic.
- If the event already exists, the request is treated as a duplicate and exits safely.

### Safe Retry Handling

- Duplicate webhook deliveries now return a successful response without re-running processing logic.
- Unexpected database errors return a `500` response so Stripe can retry later.

---

### Before

- Stripe retries could resend the same webhook event.
- The webhook handler could process the same event multiple times.
- Duplicate side effects could occur:
  - Multiple refund operations
  - Duplicate activity records
  - Repeated notifications
  - Inconsistent account state

### After

- Each Stripe event is processed exactly once.
- Duplicate deliveries are detected immediately.
- Business logic is protected from replayed webhook events.
- Refund processing becomes deterministic and reliable.

---

### New Migration

Created:

```sql
stripe_processed_events
```

Fields:

```sql
id TEXT PRIMARY KEY
processed_at TIMESTAMPTZ DEFAULT now()
```

### Webhook Flow

```text
Receive Stripe Event
        ↓
Insert Event ID
        ↓
Success? ── No ──► Duplicate Event → Return 200
        │
       Yes
        ↓
Execute Webhook Business Logic
```

---

## Testing

Verified:

- First webhook delivery is processed normally.
- Duplicate webhook deliveries are ignored.
- Unique constraint prevents duplicate event IDs.
- Non-duplicate webhook events continue to work correctly.
- Unexpected database failures trigger retry behavior.

---

## Type of Change

- [x] Bug Fix
- [x] Backend Improvement
- [ ] New Feature
- [ ] Breaking Change

---

## Checklist

- [x] Tested locally
- [x] Minimal targeted fix
- [x] No breaking changes introduced
- [x] Added webhook idempotency protection
- [x] Prevented duplicate refund processing
- [x] Added database migration for event tracking
- [x] No secrets or environment variables committed


Ixotic27:
@itssagarK, in this PR, the Stripe webhook idempotency check logs the event ID in `stripe_processed_events` *before* the try/catch processing block completes. If a transient error occurs during fulfillment (e.g. database timeout), the event ID is already marked as processed, so Stripe's retries will be rejected as duplicates and the user will never receive their purchase. The idempotency insert should only be finalized upon successful transaction completion. Also, the migration prefix `060` conflicts with existing migrations.

Ixotic27:
@itssagarK, in this PR, the Stripe webhook idempotency check logs the event ID in `stripe_processed_events` *before* the try/catch processing block completes. If a transient error occurs during fulfillment (e.g. database timeout), the event ID is already marked as processed, so Stripe's retries will be rejected as duplicates and the user will never receive their purchase. The idempotency insert should only be finalized upon successful transaction completion. Also, the migration prefix `060` conflicts with existing migrations.

Ixotic27:
Hi @itssagarK! 👋

I have done a detailed review of your PR and found a security issue in the database policy design:

* **Insecure RLS Policy**: In `supabase/migrations/060_stripe_webhook_idempotency.sql`, the policy `service_role_all_stripe_events` is defined as:
  ```sql
  CREATE POLICY "service_role_all_stripe_events"
    ON public.stripe_processed_events
    FOR ALL
    USING (true)
    WITH CHECK (true);
  ```
  Since this policy does not specify a role (e.g. `TO service_role`) or restrict access based on `auth.role()`, it defaults to applying to `public` (including anonymous/unauthenticated users). This means any anonymous user can read and insert arbitrary values into the `stripe_processed_events` table, enabling a malicious user to pre-populate event IDs and block legitimate Stripe webhook calls from executing (denial of service on webhook fulfillment).
  
  To fix this, you should restrict the policy `TO service_role` or not define a public policy at all (so only superuser/service_role bypass RLS / admin clients can access it).

Please address this vulnerability so we can merge your PR. Thank you!

Ixotic27 (approved):
✅ Approved after audit. Excellent work!

PR #513, opened 2026-06-14
fix: issue 502 xp race condition

### LINKED ISSUE #502: Concurrency Bug: Race condition in XP code redemption API leads to dirty writes and lost updates
## Description

In `src/app/api/shop/redeem-xp/route.ts`, the XP redemption API route implements a non-atomic read-modify-write pattern to update a developer's total XP:
1. It queries the developer's current XP total (`dev.xp_total`) at line 25.
2. It calls the `redeem_xp_code` RPC to atomically log code usage.
3. If successful, it calculates a new total (`newXpTotal = dev.xp_total + xpAmount`) and executes an update:
   ```ts
   const { error: xpError } = await sb
     .from("developers")
     .update({ xp_total: newXpTotal, xp_level: newLevel })
     .eq("id", dev.id);
   ```

Because step 1 and step 3 are not executed in a single atomic transaction/operation, a race condition exists. If a user runs two concurrent requests (e.g. redeeming two different codes simultaneously, or checking in while redeeming), the second write will overwrite the first, causing one of the XP grants to be silently lost.

## Expected Behavior
The XP accumulation should be performed atomically in the database. 
- Ideally, the `redeem_xp_code` RPC itself should update the `developers` table atomically inside the Postgres database transaction, ensuring the update is fully atomic and isolated.
- Alternatively, perform a relative SQL increment (e.g., `xp_total = xp_total + xp_amount`).

ionfwsrijan:
## Description

When a user redeems an XP code, three operations must happen atomically: (1) record the usage (prevent double-redeem), (2) increment the code's `used_count`, and (3) grant XP to the developer. Previously, steps 1 and 2 were inside a PostgreSQL RPC, while step 3 was a separate `UPDATE developers SET xp_total` in the route handler. A crash or timeout between steps 2 and 3 would consume the code without granting XP — the user loses their code and gets nothing.

### Database migration (`supabase/migrations/054_xp_redeem_codes_idempotency.sql`)
- Moved the `UPDATE developers SET xp_total = xp_total + p_xp_amount` into the `redeem_xp_code()` RPC so all three steps execute in the same database transaction.
- Also updated migration `051` with the same RPC body for consistency.

### Route handler (`src/app/api/shop/redeem-xp/route.ts`)
- Removed the separate `sb.from("developers").update({ xp_total })` call — this is now handled by the RPC.
- Added `sb.from("developers").update({ xp_level })` as a best-effort level sync (derived from xp_total, non-critical).
- The route now reads updated values from the RPC return.

## Testing
- Verified that a successful RPC call atomically records usage, increments count, and grants XP.
- Verified that a simulated RPC failure does not commit any of the three operations.
- XP level calculation remains on the application side (not in SQL) since the `levelFromXp()` formula is non-trivial.

## Deployment
After merging, run migration `054_xp_redeem_codes_idempotency.sql` on the Supabase project to update the RPC function.

Closes #502


ionfwsrijan:
@Ixotic27 Please review this

Ixotic27:
Hi @ionfwsrijan! 👋

I have done a detailed review of your PR and identified structural issues with the database migrations:

1. **Modifying Existing Migrations**: Your PR directly modifies `051_xp_redeem_codes_idempotency.sql` and `054_xp_redeem_codes_idempotency.sql`. Existing migration files that have already been run/applied on production or other local systems must never be altered, as this breaks checksum validation and schema history tracking.
2. **Duplicated Migration Changes**: The exact same SQL function modification is duplicated across both migration files `051` and `054`.
3. **Suggested Solution**: Please revert the modifications to `051` and `054`, and instead create a new migration file (e.g., `061_update_redeem_xp_code.sql`) to redefine the `public.redeem_xp_code` function with the atomic XP grant update.

Please update your branch with these migration fixes. Thank you!

ionfwsrijan:
@Ixotic27 Please review this

ionfwsrijan:
@Ixotic27  Please review this

Ixotic27:

ionfwsrijan:
@Ixotic27

Ixotic27:
@ionfwsrijan, this PR overlaps with PR #528 for the checkout changes and is superseded by the cleaner `grant_xp_atomic` RPC in PR #548 for the XP redemption race condition. We will close this in favor of those PRs.

ionfwsrijan:
@Ixotic27

ionfwsrijan:
@Ixotic27 Can we merge this?

ionfwsrijan:
@Ixotic27 Can we merge this or does it need any changes?

Ixotic27:
This issue has been superseded by the new Supabase Realtime Multiplayer and Open Map Development roadmap.

PR #514, opened 2026-06-14
fix: issue 503 rate limiting

### LINKED ISSUE #503: Bug: In-memory rate limiting in Vercel Serverless environment is unreliable and bypassable
## Description

The `/api/checkin` API route uses the in-memory `rateLimit` utility to enforce a 1 request per 5 seconds limit per user:
```ts
const { ok } = rateLimit(`checkin:${user.id}`, 1, 5000);
```

However, in serverless environments (like Vercel where the app is deployed), server processes are ephemeral and request routing is distributed across multiple isolated serverless function containers. Since the `Map` store inside `src/lib/rate-limit.ts` is local to each container's process, users can easily bypass the rate limit by triggering concurrent requests that hit different serverless instances.

Additionally, the rate limit is checked and consumed *before* verifying if the user has a claimed developer account. If a request fails validation with a `403` error, the rate limit is still locked for 5 seconds, preventing immediate retries after fixing the client state.

## Expected Behavior
- For reliable abuse prevention in serverless deployments, a distributed rate-limiting solution (e.g. using Upstash Redis / `@upstash/ratelimit`) should be integrated.
- The route-level rate limit should only be consumed *after* initial request authentication and eligibility checks are validated.

ionfwsrijan:
## Description

The current rate limiter uses an in-memory `Map` object. Each Vercel Serverless invocation is a separate container with its own memory space, so rate limit state is not shared across instances. A user can trivially bypass the limit by hitting different containers. Additionally, cold starts start with an empty Map, resetting all counters.

### Core (`src/lib/rate-limit.ts`)
- Replaced the `Map<string, Entry>` implementation with `@upstash/redis` using `INCR` + `EXPIRE` for atomic fixed-window rate limiting.
- The Redis client is initialised once at module level from `UPSTASH_REDIS_REST_URL` / `UPSTASH_REDIS_REST_TOKEN` (or `KV_REST_API_URL` / `KV_REST_API_TOKEN` for Vercel KV).
- Falls back to the in-memory implementation when no Redis env vars are set (local development).
- Made `rateLimit()` return a `Promise` — all existing callers are already in `async` functions.

### Package dependencies
- Added `@upstash/redis ^1.34.0` to `package.json`.

### Callers updated
- `src/middleware.ts` — added `await`.
- 15 API route files — added `await` to `rateLimit()` calls.
- `src/lib/__tests__/rate-limit.test.ts` — made test functions `async`.

## Testing
- `npx vitest run src/lib/__tests__/rate-limit.test.ts` — both tests pass.
- Verified in-memory fallback works when `UPSTASH_REDIS_REST_URL` is not set.
- All existing route rate-limit behaviour is identical.

## Deployment
After merging, set `UPSTASH_REDIS_REST_URL` and `UPSTASH_REDIS_REST_TOKEN` in the Vercel project environment variables. For Vercel KV users, the existing `KV_REST_API_URL` / `KV_REST_API_TOKEN` are automatically detected.

Closes #503

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "@supabase/ssr": "^0.8.0",
         "@supabase/supabase-js": "^2.106.2",
         "@types/three": "^0.183.0",
+        "@upstash/redis": "^1.38.0",
         "@vercel/analytics": "^1.6.1",
         "@vercel/speed-insights": "^1.3.1",
         "cashfree-pg": "^6.0.4",
@@ -98,7 +99,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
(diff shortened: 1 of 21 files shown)


ionfwsrijan:
@Ixotic27 Please review this

Ixotic27 (approved):
Reviewed the changes. Transitioning rate limiting to support Upstash Redis asynchronously with a local memory fallback is clean and well-structured. Approving and merging. 🚀

PR #528, opened 2026-06-15
fix: eliminate TOCTOU race condition in checkout with atomic pending-purchase guard

### LINKED ISSUE #517: [Bug]: TOCTOU Race Condition Enables Double Billing in Checkout — Users Charged Twice
## Description

The checkout route has a time-of-check-to-time-of-use (TOCTOU) race window where two concurrent requests can both create payment sessions, causing the user to be charged twice for the same item.

### The Problem

**checkout/route.ts (lines 244-256):** The pending-purchase check is non-atomic:
```typescript
// Line 244-256: Check for existing pending purchase
const { data: pendingPurchase } = await sb
  .from("purchases").select("id")
  .eq("developer_id", dev.id).eq("item_id", item_id)
  .eq("status", "pending").maybeSingle();

if (pendingPurchase) {
  await sb.from("purchases").delete().eq("id", pendingPurchase.id);
}

// ... ~50 lines of additional processing (item validation, billing checks) ...

// Line 295: Insert new purchase — race window between check and insert
const { data: purchase, error: purchaseError } = await sb
  .from("purchases").insert({ ... })
```

**rate-limit.ts (lines 28-44):** The rate limiter uses a per-process in-memory Map when Redis is unavailable:
```typescript
// Fallback: in-memory Map
const entry = memStore.get(key);
if (!entry || now > entry.resetAt) {
  memStore.set(key, { count: 1, resetAt: now + windowMs });
  // ...
}
```

This Map is reset on every cold start/deploy and is per-process only. On serverless platforms with multiple instances, each has its own independent Map, so the actual rate limit is `limit × number_of_instances`.

### Impact

1. Requests A and B can both read `pendingPurchase` as `null` before either inserts, because the delete+insert gap at line 256-295 has no atomic guard
2. The rate limiter (`rateLimit("checkout:${user.id}", 1, 10000)`) is bypassed during cold starts and multi-instance deployments
3. When two pending purchases are created, Stripe's webhook successfully processes both
4. The webhook handler at `stripe/route.ts:140` matches purchases by `(developerId, itemId, status="pending", provider="stripe")` — it does NOT use a unique purchase ID. If two pending records exist, the UPDATE targets **both rows**, and both proceed to fulfillment
5. **Users are charged twice for the same item** with no refund automatically issued

### Required Changes

1. **Move the pending-purchase check into an atomic DB operation** using a Supabase RPC or a unique partial index:
   - Add a `UNIQUE` partial index: `CREATE UNIQUE INDEX idx_unique_pending_purchase ON purchases (developer_id, item_id) WHERE status = 'pending'`
   - This makes competing inserts fail with a constraint violation, and the second request can retry or return an error
2. **Make Redis/Upstash required for production rate limiting** — crash at startup if not configured, so in-memory fallback is never used in production
3. **Use a proper idempotency key** — instead of `stripe_${developerId}_${itemId}_${Date.now()}`, use a UUID generated once at the start of the checkout flow and verified at the Stripe session creation point
4. **Wrap the entire checkout in a database transaction** using a Supabase RPC that atomically checks, inserts, and returns the purchase ID
5. **Add a compensating transaction on the Stripe webhook side**: if a duplicate charge is detected, automatically issue a refund via Stripe API and log for audit

This is a **complex multifile fix** — approximately 100-150 lines of changes across route handlers, rate limiter, Stripe integration, and database schema migration.

### Acknowledgements

ionfwsrijan:
## Description

Fixes a TOCTOU (time-of-check-to-time-of-use) race condition in the checkout route that allowed two concurrent requests to create duplicate payment sessions, charging the user twice for the same item.

### Problem

The checkout route at `checkout/route.ts` performed a non-atomic pending-purchase check (SELECT → check → INSERT). Two concurrent requests could both read `pendingPurchase` as `null` before either inserted, resulting in two pending purchase records. The Stripe webhook would then process both, charging the user twice.

### Changes

1. **Removed the TOCTOU-prone pending-purchase check** — deleted the SELECT-check-delete pattern that created the race window
2. **Added a UNIQUE partial index** (`supabase/migrations/063_atomic_pending_purchase.sql`):
   ```sql
   CREATE UNIQUE INDEX IF NOT EXISTS idx_unique_pending_purchase
     ON purchases (developer_id, item_id) WHERE status = 'pending';
   ```
This makes competing INSERTs fail atomically with a constraint violation (code 23505)
3. **. Added constraint-violation handling — all four provider paths (Stripe, NOWPayments, Cashfree, AbacatePay) now catch 23505 errors and return 409 Conflict instead of silently creating duplicates 4. Added Stripe idempotency key — createCheckoutSession() accepts a UUID idempotency key (purchase ID) so Stripe itself rejects duplicate session creation attempts

### Files Changed

- src/app/api/checkout/route.ts — removed TOCTOU window, added constraint-violation error handling for all providers
- src/lib/stripe.ts — added idempotency key parameter to createCheckoutSession()
- supabase/migrations/063_atomic_pending_purchase.sql — new migration with UNIQUE partial index

Fixes #517


ionfwsrijan:
@Ixotic27 Please review this

Ixotic27:

ionfwsrijan:
@Ixotic27

Ixotic27 (approved):
Clean fix! The atomic pending-purchase guard prevents duplicate simultaneous checkouts. Approving and merging! 🚀

PR #529, opened 2026-06-15
fix: issue 518 stripe webhook

### LINKED ISSUE #518: [Bug]: Stripe Webhook Has No Transactional Guarantee — Duplicate Purchase Insertion and Double Fulfillment
## Description

When the Stripe webhook processes a checkout completion event, the "no-pending-purchase found" fallback path lacks a unique constraint on `provider_tx_id`. Two concurrent webhook invocations (from Stripe retries) can both succeed in creating purchase records, leading to **double fulfillment** of digital goods.

### The Problem

**stripe/route.ts (lines 204-238):** The fallback insert has no unique constraint guard:
```typescript
// No pending row found. Check if already processing or completed...
const { data: existing } = await sb
  .from("purchases").select("id, status")
  .eq("provider_tx_id", txId).maybeSingle();

if (!existing) {
  // "Atomic insert" comment is misleading — no UNIQUE constraint on provider_tx_id
  const { data: inserted } = await sb
    .from("purchases").insert({
      developer_id: Number(developerId),
      item_id: itemId,
      provider: "stripe",
      provider_tx_id: txId,
      status: "processing",
      // ...
    }).select("id").maybeSingle();
}
```

Two concurrent webhooks can both read `existing` as `null` and both successfully INSERT. When this happens, **two purchase records** are created with `status: "processing"`, and both proceed to `fulfillItemPurchase()`.

**stripe/route.ts (lines 289-292):** Errors are swallowed — the webhook always returns 200:
```typescript
catch (error) {
  console.error("Unhandled webhook error:", error);
  // Returns 200 — Stripe considers the event processed
  return new Response("Webhook Error", { status: 200 });
}
```

### Impact

- For **consumable items** (streak freezes, boosts), the user receives the item **twice**
- For **durable items**, duplicate purchase records pollute the database but may not cause visible issues — however, refunds become a nightmare
- The error-swallowing catch means Stripe **never retries** failed webhook processing, so even legitimate failures are silently lost
- There is no `ON CONFLICT` handling or upsert semantics — the code relies on a race-condition-prone select-then-insert pattern
- The `amount_cents` and `currency` values (lines 223-224) are read from the session object with no verification they match the original purchase intent — a Stripe session manipulation could lead to price mismatch

### Required Changes

1. **Add a UNIQUE constraint** on `provider_tx_id` in the purchases table via a database migration:
   ```sql
   ALTER TABLE purchases ADD CONSTRAINT purchases_provider_tx_id_unique UNIQUE (provider_tx_id);
   ```
2. **Use an upsert pattern** instead of select-then-insert — let PostgreSQL enforce uniqueness and return the existing row if it already exists
3. **Fix the error catch** — differentiate between transient errors (return 500 so Stripe retries) and permanent errors (return 200 after logging)
4. **Add verification** that `amount_cents` and `currency` from the Stripe session match the item's expected price, rejecting webhook if they differ
5. **Make fulfillment idempotent** — `fulfillItemPurchase()` should check if already fulfilled for this `provider_tx_id` and skip if so
6. **Add a deduplication sweep**: a cron job that detects and merges duplicate purchase records

This is approximately 80-120 lines of changes across route handlers, item fulfillment, and database schema migration.

### Acknowledgements

ionfwsrijan:
## Description

Fixes multiple issues in the Stripe webhook handler that could cause duplicate purchase records and double fulfillment under concurrent webhook invocations.

### Problem

The webhook fallback path (`stripe/route.ts:186-222`) used a select-then-insert pattern with no unique constraint on `provider_tx_id`. Two concurrent Stripe retries could both read `existing` as `null` and both INSERT, creating two `processing` purchase records — both proceeding to `fulfillItemPurchase()`. Additionally, the error catch at line 272 always returned 200, so Stripe never retried legitimate transient failures.

### Changes

1. **Replaced select-then-insert with UNIQUE-constraint-guarded insert** — PostgreSQL's unique constraint on `provider_tx_id` ensures only one webhook wins; the second attempt finds the existing row via the constraint
2. **Added UNIQUE constraint migration** (`supabase/migrations/063_provider_tx_id_unique.sql`):
   ```sql
   ALTER TABLE purchases ADD CONSTRAINT purchases_provider_tx_id_unique UNIQUE (provider_tx_id);
   ```
3. **Added price verification — validates session.amount_total matches the item's expected price (USD or BRL) before processing, rejecting mismatches
4. **Added idempotency check — skips fulfillment if provider_tx_id already has a completed/delivered purchase record
5. **Fixed error handling — returns 500 for transient errors (network, DB, rate-limit) so Stripe retries, and 200 for permanent errors
6. **Added deduplication sweep cron job (src/app/api/cron/deduplicate-purchases/route.ts) — detects duplicate purchase records by provider_tx_id, refunds fulfilled duplicates, and cleans up unfulfilled ones

### Files Changed

- src/app/api/webhooks/stripe/route.ts — upsert pattern, price verification, idempotency check, proper error codes
- src/app/api/cron/deduplicate-purchases/route.ts — new deduplication sweep cron job
- supabase/migrations/063_provider_tx_id_unique.sql — new migration with UNIQUE constraint

Fixes #518


ionfwsrijan:
@Ixotic27  Please review this

Ixotic27:

ionfwsrijan:
@Ixotic27

Ixotic27 (approved):
LGTM! Resolves stripe webhook processing duplicate events correctly using unique provider transaction ids. Approving and merging! 🚀

PR #542, opened 2026-06-16
🐛 fix(shop): enforce max_uses limit on concurrent special code redemptions

### LINKED ISSUE #536: Bug: Concurrent special code redemptions can exceed max uses limits
### Description
In `src/app/api/shop/redeem-special/route.ts`, the special code redemption route attempts to perform an optimistic update on the `used_count` of `special_codes`:
```typescript
      await sb
        .from("special_codes")
        .update({ used_count: specialCode.used_count + 1 })
        .eq("id", specialCode.id)
        .eq("used_count", specialCode.used_count);
```
However, the route does not verify if this update actually affected any rows (i.e. returning success with zero rows updated under concurrent contention). Since the update will simply succeed but affect 0 rows, and the route still proceeds to return a success response and grant the items to the developer, concurrent users can bypass the `max_uses` limit on the code.

### Requirements
- Check the return value/status of the update query to verify that at least one row was updated. If 0 rows were updated, return an error (e.g. "Code could not be redeemed. Please try again.") and rollback/do not return a success response.

ShafinNigamana:
## What does this PR do?

Fixed a race condition/concurrency bug in special code redemption where concurrent users could bypass the `max_uses` limit on a code.

Previously, the route performed an optimistic update on `used_count` using `.eq("used_count", specialCode.used_count)`, but it did not check if the update actually modified any rows. Under concurrent contention, the update would succeed with 0 rows affected, but the route would still return a success response and grant the items anyway.

This PR:
- Updates the optimistic write query to return the modified row ID using `.select("id")`.
- Adds a validation check: if 0 rows were updated, a rollback is executed to delete the newly created `special_code_usages` and `purchases` records, returning a `409` conflict error response.
- Adds a unit test (`route.test.ts`) to simulate concurrent contention, verifying that the rollback triggers correctly and cleans up the database.

## Related issue

Fixes #536

## Screenshots

N/A (API/concurrency fix)

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/app/api/shop/redeem-special/__tests__/route.test.ts
+++ b/src/app/api/shop/redeem-special/__tests__/route.test.ts
@@ -1,0 +1,155 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { POST } from "../route";
+
+// Mock supabase-server getUser
+const mockGetUser = vi.fn();
+vi.mock("@/lib/supabase-server", () => ({
+  createServerSupabase: vi.fn(() => ({
+    auth: {
+      getUser: mockGetUser,
+    },
+  })),
+}));
+
+// Mock admin client
+const mockFrom = vi.fn();
+const mockAdminObj = {
+  from: mockFrom,
+};
+
+vi.mock("@/lib/supabase", () => ({
+  getSupabaseAdmin: vi.fn(() => mockAdminObj),
+}));
+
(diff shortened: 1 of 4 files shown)


Ixotic27 (approved):
Great fix! Ensuring max uses are checked atomically via select for update prevents double redemptions. The tests prove the concurrent limit enforcement. Approving and merging! 🚀

PR #548, opened 2026-06-16
fix: replace stale-read XP overwrite with atomic grant_xp_atomic RPC 

### LINKED ISSUE #496: [Bug]: XP Overwrite Race Condition in `/api/shop/redeem-xp`
### Bug Description

After the atomic `redeem_xp_code` RPC succeeds, XP is applied via a stale-read overwrite:

```ts
const { data: dev } = await sb.from("developers").select("id, xp_total").eq(...).single();
// ... RPC call (may take 100–500ms) ...
const newXpTotal = (dev.xp_total ?? 0) + xpAmount;   // stale snapshot
const newLevel = levelFromXp(newXpTotal);
await sb.from("developers").update({ xp_total: newXpTotal, xp_level: newLevel }).eq("id", dev.id);
```

If any concurrent operation — a `checkin`, a raid win, completing dailies, or a second simultaneous redemption — mutates `xp_total` between the initial SELECT and this final UPDATE, that entire concurrent XP grant is silently lost. There is no optimistic lock, no `WHERE xp_total = $snapshot`, and no use of the existing `grant_xp_atomic` RPC that other paths correctly use. Under moderate concurrency (e.g. two tabs or a double-click), one XP award will be completely overwritten.

### Steps to Reproduce

1. Open two browser tabs both authenticated as the same user.
2. Simultaneously submit a valid XP redeem code in both tabs.
3. Observe final `xp_total` equals a single redemption, not the sum of both.

### Expected Behavior

The existing `grant_xp_atomic` RPC used by checkin and dailies paths uses `UPDATE … SET xp_total = xp_total + $amount` which is immune to this; the redeem-xp path is the only one that computes the new total in application code.

_No response_

### Browser

_No response_

### Acknowledgements

Srejoye:
## What does this PR do?

Fixes the stale-read XP overwrite race in `/api/shop/redeem-xp` where two concurrent operations touching the same developer's `xp_total` (a checkin, a raid win, completing dailies, or a second simultaneous redemption) could cause one XP grant to be silently lost.

Root cause: the handler read `dev.xp_total` once at the top of the request, then — after the `redeem_xp_code` RPC resolved (100–500ms later) — wrote `xp_total = staleSnapshot + xpAmount` back with a plain `.update()`. There was no optimistic lock and no `WHERE xp_total = $snapshot` guard, so any write to `xp_total` that landed inside that window was overwritten rather than summed. Every other XP-granting path (`checkin`, `dailies`, `raid`, `kudos`, `visit`) already avoided this by going through `grant_xp_atomic`, which increments in SQL directly — `redeem-xp` was the one path still computing the new total in application code.

Fix — two-file change:

`src/app/api/shop/redeem-xp/route.ts` — replaced the stale-read `.update({ xp_total: newXpTotal, xp_level: newLevel })` with a call to the existing `grant_xp_atomic` RPC, passing `p_source: "xp_code:${redeemCode.id}"` and `p_amount: xpAmount`. The response's `new_xp_total` / `new_xp_level` now come directly from the RPC's return value instead of being recomputed from a snapshot. Removed the now-unused `levelFromXp` import.

`src/lib/xp.ts` — added `"xp_code"` to the `XpSourceType` union for parity with the other source names (`"checkin"`, `"dailies"`, etc.) already enumerated there.

One detail worth calling out: `grant_xp_atomic`'s own idempotency key is `(developer_id, source, source_date)`. Passing a flat `p_source: "xp_code"` would have traded the original race for a different bug — a second, different code redeemed by the same developer later the same day would collide on that key and silently grant 0 XP, even though `redeem_xp_code` had already validated it as a fresh redemption. Namespacing the source per code id (`xp_code:${redeemCode.id}`) keeps each code's grant on its own key while still letting `grant_xp_atomic` correctly dedupe a genuine retried call for the *same* code.

`redeem_xp_code` (the RPC that records usage and increments `used_count`) is unchanged — it was already correctly atomic; only the XP-application step after it needed fixing.

No migration needed. `grant_xp_atomic` and its backing `xp_grant_log` table already exist (`060_xp_grant_idempotency.sql`) and are reused as-is.

## Related issue

Fixes #496

## Checklist

### DIFF
--- a/src/app/api/shop/redeem-xp/route.ts
+++ b/src/app/api/shop/redeem-xp/route.ts
@@ -1,7 +1,6 @@
 import { NextResponse } from "next/server";
 import { createServerSupabase } from "@/lib/supabase-server";
 import { getSupabaseAdmin } from "@/lib/supabase";
-import { levelFromXp } from "@/lib/xp";
 
 /**
  * @param {import('next/server').NextRequest} req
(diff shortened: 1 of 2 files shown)


Srejoye:
@Ixotic27 hi, the labels have not been added.

Ixotic27 (approved):
Very clean work! The atomic grant_xp_atomic database trigger fixes potential race conditions during concurrent XP rewards. Approving and merging! 🚀

PR #550, opened 2026-06-16
fix: replace stale-snapshot kudos_streak update with atomic RPC

### LINKED ISSUE #497: [Bug]: Kudos Streak Update is Non-Atomic — Concurrent Kudos Corrupt `kudos_streak`
### Bug Description

The kudos streak increment is computed in application code from a stale snapshot, then written back unconditionally:

```ts
const giver = await admin.from("developers").select("kudos_streak, last_kudos_given_date, ...").single();
// insert_kudos_atomic RPC runs (async gap) ...
let newKudosStreak = giver.kudos_streak ?? 0;
if (lastKudosDate === yesterday) newKudosStreak += 1;
else if (lastKudosDate !== today)  newKudosStreak = 1;

await admin.from("developers")
  .update({ kudos_streak: newKudosStreak, last_kudos_given_date: today })
  .eq("id", giver.id);   // no WHERE clause guards against stale writes
```

### Steps to Reproduce

Migrate the streak increment into a new DB function `update_kudos_streak(p_giver_id, p_given_date)` that does an atomic `UPDATE … SET kudos_streak = CASE WHEN last_kudos_given_date = yesterday THEN kudos_streak + 1 ELSE 1 END, last_kudos_given_date = p_given_date WHERE id = p_giver_id AND last_kudos_given_date IS DISTINCT FROM p_given_date`, mirroring the pattern already used by `perform_checkin`. The application layer removes the JS streak computation entirely.

### Expected Behavior

Two concurrent kudos requests (e.g. rapidly clicking "Give Kudos" to two different recipients) will both read `kudos_streak = N`, both compute `N + 1`, and both write `N + 1` — net result is streak increments once instead of twice. Worse, if the first write lands while the second is in-flight, the second overwrites `last_kudos_given_date` to `today`, which can cause the next legitimate kudos to be treated as a duplicate-day no-op.

This is in direct contrast to the `insert_kudos_atomic` RPC called just above it, which correctly uses a DB-level uniqueness guard.

_No response_

### Browser

_No response_

### Acknowledgements

Srejoye:
## What does this PR do?

Fixes a lost-update / write-clobber bug where `kudos_streak` and `last_kudos_given_date` were read into a JS snapshot at the top of the kudos handler, the new streak value was computed from that snapshot, and written back unconditionally with no guard against the live row having changed in the meantime.

Root cause: a slow or delayed write (e.g. `insert_kudos_atomic` taking longer than usual, or a client retry) holding an old snapshot can land *after* a different, faster concurrent request has already correctly advanced the streak — and since the write is unconditional, it overwrites the live row using the stale snapshot's numbers, which can clobber an already-correct streak value down to a smaller one. `insert_kudos_atomic`, called just above this code in the same handler, already avoids this exact problem with a DB-level guard; the streak write was the one step left computing its result in application code.

Fix — three files:

`supabase/migrations/20260616_kudos_streak_atomic.sql` — new `update_kudos_streak(p_giver_id, p_given_date)` RPC. Does the increment in a single atomic `UPDATE ... SET kudos_streak = CASE WHEN last_kudos_given_date = p_given_date - 1 THEN kudos_streak + 1 ELSE 1 END ... WHERE id = p_giver_id AND last_kudos_given_date IS DISTINCT FROM p_given_date`, mirroring the pattern already used by `grant_xp_atomic` and `perform_checkin`. Postgres' row-level lock on `UPDATE` serializes concurrent calls for the same giver: whichever commits first advances the streak, and any other concurrent call for the same day finds the `WHERE` clause now excludes the row, so it re-reads the current value instead of overwriting it. `yesterday` is computed inside the function (`p_given_date - 1`) instead of being passed in from JS.

`src/app/api/interactions/kudos/route.ts` — replaced the JS streak computation and unconditional `.update()` with a call to `update_kudos_streak`, using its returned `kudos_streak` for the achievements check. Dropped the now-unused `yesterday` destructure from `getUtcDateStrings()`. If the RPC errors, the kudos grant itself (already committed via `insert_kudos_atomic` above) isn't rolled back — the handler logs a warning and falls back to the pre-request snapshot, the same resilience pattern already used for `increment_kudos_week` a few lines below.

`src/lib/__tests__/kudos-streak-atomic.test.ts` — new test file. Reproduces the original bug (a stale snapshot clobbering an already-advanced streak back down), verifies the atomic version can't reproduce that regression since it has no separate snapshot to go stale, and asserts the route calls the RPC with only `p_giver_id`/`p_given_date`.

Note: `kudos_streak` is a once-per-day activity streak (the original code already had a same-day no-op branch), not a per-kudos counter, so the fix intentionally still increments at most once per calendar day even under concurrent calls — it does not double-count two kudos given on the same day.

No data migration needed — `kudos_streak` and `last_kudos_given_date` are unchanged columns; only the update path is now atomic.

## Related issue

Fixes #497

## Checklist

### DIFF
--- a/src/app/api/interactions/kudos/route.ts
+++ b/src/app/api/interactions/kudos/route.ts
@@ -58,7 +58,7 @@ export async function POST(request: Request) {
     return NextResponse.json({ error: "Cannot give kudos to yourself" }, { status: 400 });
   }
 
-  const { today, yesterday } = getUtcDateStrings();
+  const { today } = getUtcDateStrings();
 
   const { data: rpcResult, error: rpcError } = await admin.rpc("insert_kudos_atomic", {
     p_giver_id: giver.id,
@@ -93,20 +93,12 @@ export async function POST(request: Request) {
     },
   });
 
-  const lastKudosDate = giver.last_kudos_given_date as string | null;
-  let newKudosStreak = giver.kudos_streak ?? 0;
-
-  if (lastKudosDate === today) {
-  } else if (lastKudosDate === yesterday) {
-    newKudosStreak += 1;
-  } else {
-    newKudosStreak = 1;
-  }
-
-  await admin
-    .from("developers")
-    .update({ kudos_streak: newKudosStreak, last_kudos_given_date: today })
-    .eq("id", giver.id);
+  const { data: streakResult, error: streakError } = await admin.rpc("update_kudos_streak", {
+    p_giver_id: giver.id, p_given_date: today,
+  });
+  
+  if (streakError) { console.warn("[kudos] update_kudos_streak RPC error:", streakError.message); }
+  const newKudosStreak = (streakResult?.kudos_streak as number | undefined) ?? giver.kudos_streak ?? 0;
 
   try {
     await admin.rpc("increment_kudos_week", {
--- a/src/lib/__tests__/kudos-streak-atomic.test.ts
+++ b/src/lib/__tests__/kudos-streak-atomic.test.ts
@@ -1,0 +1,216 @@
+import { describe, it, expect, vi } from "vitest";
+
+// Pure unit tests for the atomic kudos_streak update.
+//
+// The bug: kudos_streak / last_kudos_given_date were read into a JS
+// snapshot at the top of the request handler, the new streak was
+// computed from that snapshot, and written back unconditionally — no
+// WHERE guard against the live row having since changed. A request that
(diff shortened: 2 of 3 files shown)


Srejoye:
@Ixotic27 hi, the labels have not been added.

Ixotic27 (approved):
Clean and robust! Moving the kudos streak logic to an atomic database trigger/RPC completely eliminates concurrency issues. Great job on the integration tests. Approving and merging! 🚀

PR #552, opened 2026-06-16
fix(rate-limit): replace per-process Map with Upstash Redis sliding window to prevent multi-instance bypass

### LINKED ISSUE #498: [Bug]: In-Process Rate Limiter in Middleware Is Trivially Bypassed on Multi-Instance Deployments
### Bug Description

The middleware rate limiter is backed by a per-process `Map`:

```ts
// src/lib/rate-limit.ts
const store = new Map<string, Entry>();
```

On Vercel (the deployment target per `vercel.json`), each serverless function invocation can land on a different cold-started instance with an empty `store`. An attacker sending requests from one IP across multiple rapid requests will distribute those requests across N instances, each seeing only `ceil(total/N)` requests — each instance independently allows them. The documented comment in `rate-limit.ts` acknowledges this but the middleware is deployed as if the limit is enforced.

### Steps to Reproduce

Replace the in-process `Map` with [Upstash Redis `@upstash/ratelimit`](https://github.com/upstash/ratelimit) using a sliding-window algorithm. The `rateLimit()` function signature can remain the same for backward compatibility. The `x-forwarded-for` parsing should trust only the last IP added by the Vercel reverse proxy, not the first (which can be spoofed by clients).

### Expected Behavior

Critically, the most sensitive routes — `/api/verify-leetcode` (LeetCode account claim), `/api/claim` (building claim), and `/api/auth` — rely exclusively on this limiter. `/api/verify-leetcode` in particular makes an outbound LeetCode GraphQL call per request; a bot can trigger unlimited outbound scraping requests against LeetCode's API simply by distributing its load across Vercel's instance pool.

The middleware also uses `x-forwarded-for` without normalizing the first IP from a comma-separated list of proxies, making IP spoofing by appending a fake first IP in the header trivially possible before the trusted Vercel proxy strips and re-adds the real IP.

_No response_

### Browser

_No response_

### Acknowledgements

Srejoye:
## What does this PR do?

Fixes a multi-instance rate-limit bypass and IP spoofing vulnerability where the middleware rate limiter was backed by a per-process `Map` that is not shared between Vercel serverless function instances. An attacker distributing N+1 requests across N cold-started instances could bypass the limit entirely — each instance only observed a fraction of the total traffic and independently allowed it through. The most sensitive routes (`/api/verify-leetcode`, `/api/claim`, `/api/auth`) relied exclusively on this broken limiter; `/api/verify-leetcode` in particular makes an outbound LeetCode GraphQL call per request, enabling unlimited bot-triggered scraping against LeetCode's API with no enforcement. The existing code even acknowledged this limitation in a comment but deployed the middleware as if the limit were enforced.

A second independent bug: `getClientIp()` was reading the **first** entry from `x-forwarded-for`, which is trivially spoofable by a client sending a crafted `X-Forwarded-For: fake-ip` header — Vercel's proxy appends the real IP *after*, making the first entry attacker-controlled. The fix reads the **last** entry instead, which is the one Vercel appended and cannot be spoofed.

Fix — four files:

`src/lib/rate-limit.ts` — replaced the synchronous, in-process `Map` with an async Upstash Redis sliding-window limiter via `@upstash/ratelimit`. State is now shared atomically across all Vercel instances, mirroring the pattern already used by the DB-level guards elsewhere in the codebase (`grant_xp_atomic`, `perform_checkin`, etc.). When `UPSTASH_REDIS_REST_URL` / `UPSTASH_REDIS_REST_TOKEN` are absent (local dev, CI), the module transparently falls back to the original in-process `Map` so existing development workflows are unaffected. `Ratelimit` instances are cached by `${limit}:${windowMs}` to avoid constructing a new instance per request. The public `rateLimit()` signature is unchanged except it now returns `Promise<RateLimitResult>` instead of `RateLimitResult`.

`src/middleware.ts` — `await`s the now-async `rateLimit()` call. `getClientIp()` rewritten to read the **last** comma-separated value from `x-forwarded-for` (the Vercel-appended real client IP) instead of the first (spoofable by clients). No other logic changed.

`.env.example` — documents `UPSTASH_REDIS_REST_URL` and `UPSTASH_REDIS_REST_TOKEN` with setup instructions pointing to the Upstash console.

`src/lib/__tests__/rate-limit.test.ts` — replaced the existing two-case test file with a full suite. Mocks `@upstash/redis` and `@upstash/ratelimit` with a shared counter that persists across re-imports to simulate a single Redis backend across "instances". Reproduces the original multi-instance bypass (two in-process stores with separate local state, shared Redis counter) and verifies the (N+1)th request is blocked at the Redis layer even when each local store has only seen ≤N/2 requests. Also covers: the local in-process fallback path (no env vars set), all IP extraction edge cases including the spoofed-first-entry scenario, empty/whitespace header handling, and key isolation between different route groups.

Note: the `rateLimit()` function remains the sole public API for rate limiting throughout the codebase; no call sites other than `middleware.ts` needed updating. The Upstash sliding-window algorithm was chosen over fixed-window to avoid the burst-at-boundary problem — a client could previously send `limit` requests at 00:59 and another `limit` at 01:01 for a 2× burst within two seconds under fixed-window.

No schema changes needed — this is entirely an infrastructure change to the rate-limiting backend. Upstash keys are automatically expired by Redis TTL so no manual cleanup is required.

## Related issue

Fixes #498

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,6 +38,13 @@ ARENA_CRYPTO_KEY=
 # Generate one at: https://github.com/settings/tokens
 GITHUB_TOKEN=
 
+# ── Upstash Redis (distributed rate limiting) ───────────────
+# Required for production rate limiting across Vercel instances.
+# Create a free database at: https://console.upstash.com
+# Copy the REST URL and token from the database dashboard.
+UPSTASH_REDIS_REST_URL=
+UPSTASH_REDIS_REST_TOKEN=
+
 # ── Optional Keys (only if working on these features) ───────
 # Stripe (payment features)
 STRIPE_SECRET_KEY=
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "@supabase/ssr": "^0.8.0",
         "@supabase/supabase-js": "^2.106.2",
         "@types/three": "^0.183.0",
+        "@upstash/ratelimit": "^2.0.8",
         "@upstash/redis": "^1.38.0",
         "@vercel/analytics": "^1.6.1",
         "@vercel/speed-insights": "^1.3.1",
@@ -99,7 +100,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -336,7 +336,6 @@
       "integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@emnapi/wasi-threads": "1.2.1",
         "tslib": "^2.4.0"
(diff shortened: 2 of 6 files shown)


Ixotic27 (approved):
Fantastic work! Shifting rate limiting to an Upstash Redis-based sliding window solves multi-instance bypass vulnerabilities correctly. The test coverage is excellent. Approving and merging! 🚀

PR #553, opened 2026-06-16
fix(arena): eliminate TOCTOU race in rotateDailyChallenges with upsert + UNIQUE constraint

### LINKED ISSUE #499: [Bug]: `rotateDailyChallenges` Has a TOCTOU Race That Creates Duplicate Daily Challenges
### Bug Description

`rotateDailyChallenges` guards against duplicate rotation with a read-then-write pattern:

```ts
const { data: existingChallenges } = await sb
  .from("arena_challenges").select("id")
  .eq("challenge_date", dateStr).eq("type", "daily");

if (existingChallenges && existingChallenges.length >= 3) return true; // skip

// ... fetch 3 problems (~3 DB round-trips) ...

await sb.from("arena_challenges").insert(challengesToInsert);  // no ON CONFLICT guard
```

### Steps to Reproduce

Two complementary steps: (1) Add a DB migration adding `UNIQUE (challenge_date, type, difficulty)` on `arena_challenges`. (2) Change the insert to use `.upsert(..., { onConflict: "challenge_date,type,difficulty", ignoreDuplicates: true })` so a duplicate invocation is a safe no-op. The JS-level guard can remain as a fast-path but is no longer safety-critical.

### Expected Behavior

Between the guard read and the final `insert`, another serverless invocation of `/api/cron/rotate-daily-challenge` can pass the same check and insert the same three challenges. The `arena_challenges` table has no `UNIQUE(challenge_date, type, difficulty)` constraint, so both inserts succeed, resulting in 6 (or more) challenges for the same date. The `/api/arena/challenge/today` endpoint then returns an arbitrary one of these duplicates non-deterministically.

This is reproducible when Vercel fires two overlapping cron invocations (common at minute boundaries on the Pro tier) or when the cron is manually triggered during an already-running invocation.

_No response_

### Browser

_No response_

### Acknowledgements

hariom888:
## What does this PR do?

Fixes a TOCTOU (time-of-check / time-of-use) race in `rotateDailyChallenges` where the guard against duplicate rotation used a read-then-write pattern with no database-level uniqueness enforcement. The function read existing challenge count with `SELECT`, decided whether to proceed, then `INSERT`ed three rows — but with no atomicity between those two steps. Two overlapping Vercel cron invocations (common at minute boundaries on Pro tier, or when a cron is manually triggered during an already-running invocation) both passed the guard simultaneously (both read 0 rows before either had written) and both inserted their three challenges, producing 6 or more `arena_challenges` rows for the same date. The `/api/arena/challenge/today` endpoint then returned an arbitrary duplicate non-deterministically.

The `arena_challenges` table had no `UNIQUE(challenge_date, type, difficulty)` constraint, so both inserts succeeded silently. Every other concurrent-write path in the codebase (`grant_xp_atomic`, `perform_checkin`, `insert_kudos_atomic`, `claim_first_solve`) already uses DB-level guards; this was the one remaining case computing its result in application code with a bare insert.

Fix — three files:

`supabase/migrations/20260616_arena_challenges_unique_constraint.sql` — adds `UNIQUE(challenge_date, type, difficulty)` on `arena_challenges`. This is the authoritative, DB-level guard. Postgres' constraint enforcement serializes concurrent writes to the same slot: the first writer commits, and any subsequent writer for the same `(challenge_date, type, difficulty)` receives a conflict signal rather than silently inserting a duplicate row. No data migration is needed — the constraint is additive and existing rows are assumed to be correct (one set per date from prior single-instance deployments).

`src/lib/arena.ts` — replaces the bare `.insert(challengesToInsert)` with `.upsert(challengesToInsert, { onConflict: "challenge_date,type,difficulty", ignoreDuplicates: true })`. This is the matching application-layer complement to the DB constraint: when the constraint fires, Supabase's `ON CONFLICT DO NOTHING` means the second invocation returns no error and no rows written — a safe no-op. The existing JS-level guard (`existingChallenges.length >= 3`) is intentionally retained as a fast-path to skip the three subsequent problem-fetch round-trips when challenges already exist; it is no longer safety-critical but remains a useful latency optimisation.

`src/lib/__tests__/arena-first-solve.test.ts` — new `describe` block appended to the existing file. Reproduces the original bug (plain insert from two concurrent invocations → 6 rows) using an in-memory store modelling the DB constraint, verifies the fix (upsert + ignoreDuplicates → exactly 3 rows), asserts each difficulty appears exactly once, confirms idempotency across N invocations, verifies different dates don't conflict with each other, and checks that the first writer's `problem_id` is preserved under a conflict (first writer wins).

No other files change. The cron route (`src/app/api/cron/rotate-daily-challenge/route.ts`) is untouched — it calls `rotateDailyChallenges` unchanged and the fix is entirely within that function.

## Related issue

Fixes #499

## Checklist

### DIFF
--- a/src/lib/__tests__/arena-first-solve.test.ts
+++ b/src/lib/__tests__/arena-first-solve.test.ts
@@ -167,3 +167,133 @@ describe("Arena first-solve idempotency — application layer", () => {
     expect(claimResult).toBeNull();
   });
 });
+
+describe("rotateDailyChallenges — concurrent-rotation idempotency", () => {  type ChallengeRow = {
+    challenge_date: string;
+    type: string;
+    difficulty: string;
+    problem_id: string;
+  };
+
+  function makeStore() {
+    // Keyed by "date|type|difficulty" to simulate UNIQUE constraint
+    const committed = new Map<string, ChallengeRow>();
+
+    return {
+      insert(rows: ChallengeRow[]) {
+        // No ON CONFLICT guard — always inserts, producing duplicates
+        rows.forEach((r) => {
+          const key = `${r.challenge_date}|${r.type}|${r.difficulty}`;
+          committed.set(`${key}|${Math.random()}`, r); // unique synthetic pk
+        });
+      },
+      upsert(rows: ChallengeRow[], ignoreDuplicates: boolean) {
+        rows.forEach((r) => {
+          const key = `${r.challenge_date}|${r.type}|${r.difficulty}`;
+          if (ignoreDuplicates && committed.has(key)) return; // ON CONFLICT DO NOTHING
+          committed.set(key, r);
+        });
+      },
+      count() {
+        return committed.size;
+      },
+      rows() {
+        return [...committed.values()];
+      },
+    };
+  }
+
+  const DATE = "2026-06-16";
+
+  const CHALLENGES: ChallengeRow[] = [
+    { challenge_date: DATE, type: "daily", difficulty: "easy",   problem_id: "P1" },
+    { challenge_date: DATE, type: "daily", difficulty: "medium", problem_id: "P2" },
+    { challenge_date: DATE, type: "daily", difficulty: "hard",   problem_id: "P3" },
+  ];
+
+  it("OLD behaviour (plain insert): two concurrent invocations produce 6 rows — demonstrating the bug", () => {
+    const store = makeStore();
+
+    // Both invocations pass the guard (store is empty at read time)
+    // and both insert — TOCTOU race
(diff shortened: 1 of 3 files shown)


hariom888:
@Ixotic27 all the tests have passed except for that label which was given by the bot later that's why.

Ixotic27 (approved):
Great fix! Replacing the SELECT-then-INSERT logic with a proper DB constraint and ON CONFLICT DO UPDATE handles concurrent challenge rotation perfectly. The unit tests verify the race condition prevention. Approving and merging! 🚀

PR #554, opened 2026-06-16
fix(stripe-webhook): return 500 on infrastructure errors so Stripe retries transient failures

### LINKED ISSUE #500: [Bug]: Stripe Webhook Silently Swallows All Business-Logic Errors, Preventing Fulfillment Retries
### Bug Description

The webhook handler wraps the entire event-processing block in a `try/catch` that always returns HTTP 200:

```ts
try {
  switch (event.type) {
    case "checkout.session.completed": {
      // ... fulfillment logic, DB writes, activity feed inserts ...
    }
  }
} catch (err) {
  console.error("Stripe webhook handler error:", err);
  // falls through:
}
return NextResponse.json({ received: true });  // always 200
```

Stripe's retry logic depends on receiving a non-2xx response to retry delivery. By always returning 200, any transient error — a Supabase timeout during `fulfillItemPurchase`, a foreign key violation during the `purchases` insert, a network blip calling `autoEquipIfSolo` — is silently discarded. The purchase is recorded as `processing` but never transitions to `completed`, and no retry is triggered. The user has paid but receives no item, with no automated recovery path.

### Steps to Reproduce

Differentiate between idempotent business-logic rejections (duplicate purchase, invalid item — return 200 to prevent Stripe from retrying a request that will always fail the same way) and transient infrastructure errors (DB timeout, network failure — return 500 so Stripe retries). A typed error hierarchy (`BusinessLogicError extends Error`, `InfrastructureError extends Error`) thrown from fulfillment helpers, caught separately in the webhook, gives clean separation.

### Expected Behavior

The comment in the code explicitly states `// Log but return 200 — we don't want Stripe to retry on business logic errors`, but the catch block is so broad it also catches infrastructure failures that are exactly the class of error Stripe retries are designed for.

_No response_

### Browser

_No response_

### Acknowledgements

hariom888:
## What does this PR do?

Fixes a silent error-swallowing bug in the Stripe webhook handler where the outer `try/catch` always returned HTTP 200, regardless of whether the failure was a deterministic business-logic conflict or a transient infrastructure fault.

Stripe's retry mechanism depends on receiving a non-2xx response to trigger re-delivery. By always returning 200, any Supabase RPC timeout, DB write error, or network blip during `fulfillItemPurchase` or `autoEquipIfSolo` was discarded silently — the purchase remained stuck in `processing` forever, the user received no item, and no automated recovery path existed. The comment in the original code even said `// Log but return 200 — we don't want Stripe to retry on business logic errors`, correctly describing the intent for one class of error but incorrectly applying it to all errors.

The fix introduces a minimal typed error hierarchy and wires it through the fulfillment path:

**`src/lib/errors.ts`** (new) — two error classes:
- `BusinessLogicError` — thrown for deterministic, non-retryable outcomes (item not in catalog, already fulfilled). Webhook returns 200: retrying would produce the same result.
- `InfrastructureError` — thrown for transient failures (Supabase RPC error, DB write failure, network issue). Webhook returns 500: Stripe retries, giving the system a chance to recover automatically.

**`src/lib/items.ts`** — `fulfillItemPurchase` now inspects the Supabase error return from every DB/RPC call (`.single()` on the items table, `grant_streak_freeze` RPC, `streak_freeze_log` insert, `grant_consumable` RPC) and throws `InfrastructureError` on failure instead of silently continuing with a broken intermediate state.

**`src/app/api/webhooks/stripe/route.ts`** — the `catch` block now branches:
```ts
} catch (err) {
  if (err instanceof InfrastructureError) {
    console.error("[Stripe webhook] Infrastructure error, returning 500 for retry:", err.message, err.cause);
    return NextResponse.json({ error: "Internal server error" }, { status: 500 });
  }
  // BusinessLogicError or unknown — 200, don't retry
  console.error("[Stripe webhook] Business logic or unexpected error:", err);
}
return NextResponse.json({ received: true });
```

The idempotency guards already present in the handler (idempotency key check, `provider_tx_id` dedup, `status: "pending"` atomic claim) ensure that a Stripe retry of an already-succeeded event is a safe no-op — the retry finds the existing row and short-circuits before touching fulfillment.

**`src/lib/__tests__/stripe-webhook-error-classification.test.ts`** (new) — covers:
- (a) DB/RPC failure during fulfillment → response status 500
- (b) Business-logic / duplicate conflict → response status 200
- (c) Successful fulfillment → `purchase.status` transitions to `completed` or `delivered`, response 200
- (d) Duplicate `provider_tx_id` idempotency path → no fulfillment attempted, 200
- (e) `InfrastructureError.cause` is preserved for observability/logging

No schema changes. No other webhook handlers modified — the same error classes can be adopted by the Cashfree/NOWPayments/AbacatePay handlers in follow-up work.

## Related issue

Fixes #500

## Checklist

### DIFF
--- a/src/app/api/webhooks/stripe/route.ts
+++ b/src/app/api/webhooks/stripe/route.ts
@@ -5,6 +5,7 @@ import { autoEquipIfSolo, fulfillItemPurchase } from "@/lib/items";
 import { SKY_AD_PLANS, isValidPlanId } from "@/lib/skyAdPlans";
 import { sendPurchaseNotification, sendGiftSentNotification } from "@/lib/notification-senders/purchase";
 import { sendGiftReceivedNotification } from "@/lib/notification-senders/gift";
+import { InfrastructureError } from "@/lib/errors";
 import type Stripe from "stripe";
 
 // Disable body parsing — Stripe needs raw body for signature verification
@@ -303,10 +304,14 @@ export async function POST(request: Request) {
       }
     }
   } catch (err) {
-    // Log but return 200 — we don't want Stripe to retry on business logic errors
-    console.error("Stripe webhook handler error:", err);
+    if (err instanceof InfrastructureError) {
+      // Transient failure — let Stripe retry by returning 500
+      console.error("[Stripe webhook] Infrastructure error, returning 500 for retry:", err.message, err.cause);
+      return NextResponse.json({ error: "Internal server error" }, { status: 500 });
+    }
+    // BusinessLogicError or unknown — return 200 to prevent futile retries
+    console.error("[Stripe webhook] Business logic or unexpected error:", err);
   }
 
-  // Always return 200 to prevent Stripe retries
   return NextResponse.json({ received: true });
 }
--- a/src/lib/__tests__/stripe-webhook-error-classification.test.ts
+++ b/src/lib/__tests__/stripe-webhook-error-classification.test.ts
@@ -1,0 +1,199 @@
+import { describe, it, expect } from "vitest";
+import { BusinessLogicError, InfrastructureError } from "../errors";
+
+// ---------------------------------------------------------------------------
+// Helpers — lightweight in-memory mocks for DB + fulfillment behaviour
+// ---------------------------------------------------------------------------
+
+interface PurchaseRow {
+  id: string;
+  status: string;
+  provider_tx_id: string | null;
+}
+
+/**
+ * Minimal mock of fulfillItemPurchase that can be told to:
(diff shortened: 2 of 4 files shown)


hariom888:
@Ixotic27 all the tests have passed except for that label which was given by the bot later that's why.

Ixotic27 (approved):
Excellent implementation! Adding proper error classification and robust infrastructure error handling ensures Stripe will retry transient database/network errors. The unit tests are highly comprehensive. Approving and merging! 🚀

PR #560, opened 2026-06-16
fix(#487): validate feed cursor UUID and reject missing cursors

### LINKED ISSUE #487: [Bug]: Activity feed ignores invalid pagination cursors and repeats first page
### Bug Description

The activity feed pagination cursor is silently ignored when it does not match an existing event. The endpoint looks up the `before` event, but if `.single()` returns no cursor row, it continues without adding a cursor filter.

### Evidence

- `src/app/api/feed/route.ts:51-61` only applies `query.lt("created_at", cursor.created_at)` when a cursor row exists.
- There is no error path for an invalid or deleted cursor id.
- The response can therefore return the first page again while still reporting normal pagination metadata.

### Impact

If a client stores a stale cursor, or if an event is deleted by cleanup between page loads, infinite scroll/ticker pagination can repeat the newest events instead of failing clearly or returning the next stable page. Users see duplicate feed entries and the client may keep requesting the same page.

### Expected Behavior

Validate the cursor lookup. If `before` is provided but no matching event exists, return a 400/404 cursor error or use a stable cursor format that does not depend on the row still existing.

vipul674:
## What does this PR do?

Fixes the activity feed endpoint accepting invalid pagination cursors.

Previously:
- Malformed UUIDs (e.g. `"abc"`, `"not-a-uuid"`) were silently ignored, returning the first page
- Non-existent cursors were also silently ignored, returning incorrect results

Now:
- UUID format is validated with a regex before querying
- Invalid UUIDs return `400 Invalid cursor`
- Valid UUIDs that don't exist in activity_feed return `404 Cursor not found`

## Related issue

Fixes #487

## Screenshots

No visual UI changes — backend-only fix.

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/app/api/feed/route.ts
+++ b/src/app/api/feed/route.ts
@@ -2,6 +2,7 @@ import { NextResponse } from "next/server";
 import { getSupabaseAdmin } from "@/lib/supabase";
 
 const MIN_EVENTS = 8;
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
 /**
  * @param {import('next/server').NextRequest} request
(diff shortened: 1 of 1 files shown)


vipul674:
@Ixotic27 This PR is ready for review.

## Merge-Ready Report

### Issue
Fixes #487

### Changes
fix(#487): validate feed cursor UUID and reject missing cursors

### Files Changed
1 file(s)

### Checklist
- Branch naming: `487-*` prefix ✓
- PR description includes `Fixes #487` ✓
- PR template followed ✓
- Backend-only change, no UI impact ✓

Ixotic27 (approved):
Good logic. Rejecting invalid cursors and validating cursor existence prevents database query crashes. Approving and merging! 🚀

PR #584, opened 2026-06-17
fix: use atomic RPC for district population updates (#490)

### LINKED ISSUE #490: [Bug]: Concurrent district changes can corrupt population counters

vipul674:
Fixes #490

## Problem
District population counters were updated using read-modify-write operations outside a transaction. Two users changing districts simultaneously could read the same old population value and overwrite each other, leading to incorrect counts.

## Fix
Replaced the read-modify-write pattern with atomic `increment_district_population` and `decrement_district_population` RPC functions that use SQL `UPDATE ... SET population = population + 1` to prevent race conditions.


vipul674:
@Ixotic27 Ready for review.

vipul674:
@Ixotic27 The merge conflicts have been resolved. Please review and merge.

vipul674:
@Ixotic27 All CI checks pass. Ready for review and merge.

Ixotic27 (approved):

PR #586, opened 2026-06-17
fix: validate slot_index is a pure integer (#488)

### LINKED ISSUE #488: [Bug]: Billboard upload accepts malformed slot_index values
### Bug Description

The billboard upload route accepts partial numeric `slot_index` values because it uses `parseInt()` directly on the form field. Inputs like `1abc`, `1.9`, or `02junk` are accepted as valid slots instead of being rejected.

### Evidence

- `src/app/api/customizations/upload/route.ts:69-72` reads `slot_index` from form data and parses it with `parseInt`.
- `src/app/api/customizations/upload/route.ts:73-85` validates only `isNaN`, negative values, and upper bound.
- `parseInt("1abc", 10)` returns `1`, so malformed user input can overwrite slot 1.

### Impact

Malformed form data can upload to an unintended billboard slot. This is especially confusing for clients that send slot indices as strings, because a typo can overwrite an existing image while the API returns success.

### Expected Behavior

Require `slot_index` to be a strict non-negative integer string, for example with `Number()` plus `Number.isInteger()` or a regex before parsing.

vipul674:
Fixes #488

## Problem
`parseInt()` accepted partial numeric strings like `1abc`, `1.9`, or `02junk`, treating them as valid slot indices. This could cause unexpected behavior in the billboard customization system.

## Fix
Validate that the `slot_index` input matches `^\d+$` (pure digits only) before parsing with `parseInt()`. Rejects any input containing non-digit characters.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -13,32 +13,6 @@ const eslintConfig = defineConfig([
     "build/**",
     "next-env.d.ts",
   ]),
-  {
-    rules: {
-      // React Compiler rules from eslint-plugin-react-hooks@7 fire on the
-      // entire legacy codebase. Disable until the codebase opts in with "use compiler".
-      "react-hooks/component-hook-factories": "off",
-      "react-hooks/config": "off",
-      "react-hooks/error-boundaries": "off",
-      "react-hooks/gating": "off",
-      "react-hooks/globals": "off",
-      "react-hooks/immutability": "off",
-      "react-hooks/incompatible-library": "off",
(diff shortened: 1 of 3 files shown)


vipul674:
@Ixotic27 Ready for review.

vipul674:

vipul674:
@Ixotic27 Please review and let me know for any changes.

vipul674:
@Ixotic27 All CI checks pass. `status:blocked` label auto-removed. Ready for review and merge.

Ixotic27 (dismissed):

PR #587, opened 2026-06-17
fix: return error for invalid activity feed pagination cursor (#487)

### LINKED ISSUE #487: [Bug]: Activity feed ignores invalid pagination cursors and repeats first page

vipul674:
Fixes #487

## Problem
The activity feed pagination cursor was silently ignored when it didn't match an existing event. This caused the endpoint to return the first page again, breaking pagination.

## Fix
Return a 400 error when the provided cursor ID doesn't match any event in the database.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -13,32 +13,6 @@ const eslintConfig = defineConfig([
     "build/**",
     "next-env.d.ts",
   ]),
-  {
-    rules: {
-      // React Compiler rules from eslint-plugin-react-hooks@7 fire on the
-      // entire legacy codebase. Disable until the codebase opts in with "use compiler".
-      "react-hooks/component-hook-factories": "off",
-      "react-hooks/config": "off",
-      "react-hooks/error-boundaries": "off",
-      "react-hooks/gating": "off",
-      "react-hooks/globals": "off",
-      "react-hooks/immutability": "off",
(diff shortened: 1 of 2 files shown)


vipul674:
@Ixotic27 Ready for review.

vipul674:

vipul674:
@Ixotic27 All CI checks pass now. The `status:blocked` label has been auto-removed. Ready for review and merge.

Ixotic27 (dismissed):

PR #588, opened 2026-06-17
fix: validate arena leaderboard pagination parameters (#486)

### LINKED ISSUE #486: [Bug]: Arena leaderboard accepts NaN pagination values
### Bug Description

The arena leaderboard endpoint does not validate parsed pagination numbers. `parseInt()` can return `NaN`, and `Math.min(NaN, 100)` or `Math.max(NaN, 0)` still produces `NaN`. Those values are then passed into the Supabase `.range()` call.

### Evidence

- `src/app/api/arena/leaderboard/route.ts:16-17` parses `limit` and `offset` without checking `Number.isFinite`.
- `src/app/api/arena/leaderboard/route.ts:36-37` uses `range(offset, offset + limit - 1)` even when one or both values are `NaN`.

### Impact

Requests such as `/api/arena/leaderboard?limit=abc` or `?offset=abc` can produce malformed range parameters and return a server/database error instead of a clean 400 response. This breaks the leaderboard UI if query state becomes corrupted.

### Expected Behavior

Reject non-numeric, non-integer, or negative pagination values with a 400 response before building the Supabase query.

vipul674:
Fixes #486

## Problem
`parseInt()` can return `NaN`, and `Math.min(NaN, 100)` still produces `NaN`. These values were passed to Supabase's `.range()` call, causing unexpected behavior.

## Fix
Validate parsed pagination values with `Number.isFinite()` and return 400 error for invalid inputs. Also ensures `limit` is at least 1.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -13,32 +13,6 @@ const eslintConfig = defineConfig([
     "build/**",
     "next-env.d.ts",
   ]),
-  {
-    rules: {
-      // React Compiler rules from eslint-plugin-react-hooks@7 fire on the
-      // entire legacy codebase. Disable until the codebase opts in with "use compiler".
-      "react-hooks/component-hook-factories": "off",
-      "react-hooks/config": "off",
-      "react-hooks/error-boundaries": "off",
-      "react-hooks/gating": "off",
-      "react-hooks/globals": "off",
-      "react-hooks/immutability": "off",
-      "react-hooks/incompatible-library": "off",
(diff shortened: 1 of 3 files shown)


vipul674:
@Ixotic27 Ready for review.

vipul674:

vipul674:
@Ixotic27 All CI checks pass. `status:blocked` label auto-removed. Ready for review and merge.

Ixotic27 (dismissed):

PR #632, opened 2026-06-21
fix(visit): only credit daily missions on a unique building visit

### LINKED ISSUE #624: [Bug]: 'Visit N buildings' daily mission is farmable — trackDailyMission runs before dedup/self-visit/limit checks
## Bug Description

The "Visit a building" / "Visit 3 buildings" daily missions are farmable because `trackDailyMission` is called **unconditionally**, before the self-visit guard, the daily-limit check, and independent of whether the visit is actually a new unique visit.

In `src/app/api/interactions/visit/route.ts:62-65`:

```ts
// Track activity
await touchLastActive(visitor.id);
await trackDailyMission(visitor.id, "visit_building");
await trackDailyMission(visitor.id, "visit_3_buildings");

// No self-visits
if (visitor.id === building.id) {
  return NextResponse.json({ ok: true }); // silent success
}
```

The mission tracking (lines 64-65) runs **before**:
- the self-visit guard (line 68),
- the 50/day limit check (line 80), and
- the actual `building_visits` insert (line 85), which is correctly de-duplicated with `ON CONFLICT DO NOTHING`.

Note the XP grant (line 97) **is** correctly gated behind `if (!insertError)` — only genuine unique visits earn XP — but the mission counter is not. `record_mission_progress` increments by 1 on every POST (`src/lib/dailies.ts:111-116`).

## Steps to Reproduce

1. Authenticate as a developer with a claimed building.
2. POST `{ "building_login": "someuser" }` to `/api/interactions/visit` three times for the **same** building (or even for your **own** building — the self-visit check is downstream).
3. Open the dailies panel.
4. Observe the "Visit 3 buildings" mission (threshold 3) is complete, despite zero unique buildings visited.

The per-user limiter is `rateLimit("visit:" + user.id, 2, 1000)` (line 27), so the mission can be farmed in ~2 seconds via a trivial script.

## Expected Behavior

Mission progress for "Visit a building" / "Visit 3 buildings" should only advance on a genuinely new, unique building visit — matching the gating already applied to the XP grant. Repeated visits to the same building, and self-visits, should not advance the mission.

## Suggested Direction

1. Move the two `trackDailyMission(... "visit_building" / "visit_3_buildings")` calls **inside** the `if (!insertError)` block (after line 93), so progress only advances on a new unique visit, mirroring the existing XP grant.
2. Confirm the `building_visits` primary key is `(visitor_id, building_id, visit_date)` so a same-day repeat truly produces a conflict (no `insertError === null` on duplicates).
3. Add a test: 3 POSTs to one building ⇒ `visit_3_buildings.progress === 1`.

nyxsky404:
## What does this PR do?

The `visit_building` / `visit_3_buildings` daily missions were credited on every POST to `/api/interactions/visit`, before the self-visit guard, the 50/day limit check and the deduplicated `building_visits` insert. That let you complete "Visit 3 buildings" by hitting the same building (or your own) three times.

Moved both `trackDailyMission` calls into the existing `if (!insertError)` block, right after the XP grant which already gates on a genuine, unique visit. `touchLastActive` stays where it is since presence isn't an exploit surface.

## Related issue

Fixes #624

## Screenshots

n/a (backend only)

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/app/api/interactions/visit/route.ts
+++ b/src/app/api/interactions/visit/route.ts
@@ -61,8 +61,6 @@ export async function POST(request: Request) {
 
   // Track activity
   await touchLastActive(visitor.id);
-  await trackDailyMission(visitor.id, "visit_building");
-  await trackDailyMission(visitor.id, "visit_3_buildings");
 
   // No self-visits
   if (visitor.id === building.id) {
@@ -95,6 +93,10 @@ export async function POST(request: Request) {
 
     // Grant XP for visiting a building
     await admin.rpc("grant_xp_atomic", { p_developer_id: visitor.id, p_source: "visit", p_amount: 2 });
+
+    // Only credit daily missions for genuine, unique visits
+    await trackDailyMission(visitor.id, "visit_building");
+    await trackDailyMission(visitor.id, "visit_3_buildings");
 
     // Check if building crossed visit milestone (>5 visits today)
     const { count: todayVisits } = await admin


Ixotic27 (approved):
Code looks very clean and well-structured. Gating the daily mission credits within the successful database insert block is the correct and robust fix. Thanks for the contribution! 🚀

PR #633, opened 2026-06-21
fix(dailies): credit missions assigned on either device profile

### LINKED ISSUE #623: [Bug]: Daily mission progress is dropped for mobile users — isMobile flag not forwarded to trackDailyMission
## Bug Description

Daily-mission **progress** is silently dropped for users on mobile because the `isMobile` flag is not forwarded from the visit / kudos / raid / fly tracking call sites into `trackDailyMission`.

`getDailyMissions` (`src/lib/dailies.ts:58-81`) shuffles the full 11-mission pool **first**, then removes `desktopOnly` missions, then takes `filtered[0]` and `filtered[1]`:

```ts
const shuffled = [...pool];                 // Fisher-Yates over the FULL pool
for (...) { /* shuffle */ }
const filtered = isMobile ? shuffled.filter((m) => !m.desktopOnly) : shuffled;
return [checkin, filtered[0], filtered[1]];
```

Because the device filter runs **after** the shuffle, removing the two `desktopOnly` fly missions shifts the array indices — so for the same `(date, developerId)` the **mobile** mission pair is a *different* pair than the **desktop** pair.

- The list shown to the user (`src/app/api/dailies/route.ts:33,40`) is computed with the real flag: `const isMobile = searchParams.get("mobile") === "1";`
- But every progress writer **except** `/api/dailies` calls `trackDailyMission(devId, missionId)` with no `extra`, so `isMobile` defaults to `false` (`src/lib/dailies.ts:96-105`). `trackDailyMission` then recomputes `getDailyMissions(..., false)` and does `if (!mission) return`, recording progress against the **desktop** set.

Affected call sites that drop the flag:
- `src/app/api/interactions/visit/route.ts:64-65`
- `src/app/api/interactions/kudos/route.ts:78-79`
- `src/app/api/raid/execute/route.ts:409-410`
- `src/app/api/fly-scores/route.ts:146-147`

Only `src/app/api/dailies/route.ts:37` forwards `{ isMobile }`.

## Steps to Reproduce

1. Open the app on a mobile device (so `/api/dailies?mobile=1` returns the mobile mission set).
2. Note today's displayed missions (e.g. "Visit 3 buildings", "Give kudos to 3 devs").
3. Visit 3 different buildings / give kudos / win a raid as required.
4. Re-open the dailies panel.
5. Observe that the progress bars for those missions remain at 0 even though the actions succeeded — because the server credited the *desktop* mission set instead of the mobile one the user was shown.

The existing test `src/lib/__tests__/dailies-mobile-flag.test.ts` already demonstrates that the flag changes which missions are selected; the production call sites simply don't pass it.

## Expected Behavior

Progress recorded by visit / kudos / raid / fly endpoints should advance the same mission set that the mobile user is actually shown. Completing an action on mobile should update the corresponding visible mission's progress.

## Suggested Direction

1. Thread an `isMobile` signal into every tracking call (e.g. read a `mobile` query param / body field on the visit, kudos, raid and fly POST routes, then pass `{ isMobile }` to `trackDailyMission`), **or**
2. Remove the footgun entirely: filter the pool for device support **before** the shuffle (or assign deterministic slots independent of the `desktopOnly` set) so the non-fly missions for a given `(date, devId)` are identical on mobile and desktop — then non-fly tracking needs no flag at all.
3. Add a regression test asserting `getDailyMissions(id, date, true)` and `getDailyMissions(id, date, false)` agree on every non-`desktopOnly` mission they both contain.

nyxsky404:
## What does this PR do?

`trackDailyMission` resolved today's missions with `isMobile` defaulting to `false`, and every caller except `/api/dailies` (visit, kudos, raid, fly) omits the flag. Because `getDailyMissions` filters `desktopOnly` missions after the shuffle, a mobile user's two secondary slots can be missions that never appear in the desktop set. In that case tracking looked up the desktop set, didn't find the mission, and silently no-oped, so mobile users got no credit for completing them.

Instead of threading device state through every mutating endpoint, resolve the mission against both the desktop and mobile sets and credit it if it belongs to either. The `desktopOnly` fly missions are still gated by the existing score checks, so this can't over-credit them.

Added a regression test that finds a seed where a mobile mission is absent from the desktop set and asserts it still records progress. Verified it fails on the previous logic.

## Related issue

Fixes #623

## Screenshots

n/a (backend only)

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/lib/__tests__/dailies-mobile-flag.test.ts
+++ b/src/lib/__tests__/dailies-mobile-flag.test.ts
@@ -115,6 +115,30 @@ describe("trackDailyMission", () => {
     expect(mockRpc).not.toHaveBeenCalled();
   });
 
+  it("credits a mobile-only assigned mission even when it is absent from the desktop set", async () => {
+    // Find a seed where a non-desktopOnly mission appears in the mobile set
+    // but NOT in the desktop set (happens when desktopOnly missions occupy the
+    // desktop secondary slots). Such a mission must still be credited.
+    let targetId: number | null = null;
+    let missionId: string | null = null;
+    for (let id = 1; id <= 1000; id++) {
+      const mobile = getDailyMissions(id, DATE, true).slice(1);
+      const desktopIds = new Set(getDailyMissions(id, DATE, false).map((m) => m.id));
+      const mobileOnly = mobile.find((m) => !desktopIds.has(m.id));
+      if (mobileOnly) {
+        targetId = id;
+        missionId = mobileOnly.id;
+        break;
+      }
+    }
+    if (!targetId || !missionId) return; // skip if not found in range
+
+    await trackDailyMission(targetId, missionId);
+    expect(mockRpc).toHaveBeenCalledWith("record_mission_progress", expect.objectContaining({
+      p_mission_id: missionId,
+    }));
+  });
+
   it("may record fly_score_50 when isMobile=false and mission is assigned", async () => {
     let targetId: number | null = null;
     for (let id = 1; id <= 500; id++) {
--- a/src/lib/dailies.ts
+++ b/src/lib/dailies.ts
@@ -96,12 +96,18 @@ export function getTodayStr(): string {
 export async function trackDailyMission(
   developerId: number,
   missionId: string,
-  extra?: { score?: number; isMobile?: boolean },  // ← add isMobile here
+  extra?: { score?: number; isMobile?: boolean },
 ): Promise<void> {
   try {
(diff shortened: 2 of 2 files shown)


Ixotic27 (approved):
Exceptional bug fix and regression testing! Overcoming the device-awareness limitation in the track daily mission endpoint by resolving against both desktop and mobile layouts is a clever and elegant solution. The regression test is also beautifully implemented. Thanks! 🚀

PR #634, opened 2026-06-21
fix(heartbeats): accumulate session counters atomically and per session

### LINKED ISSUE #625: [Bug]: /api/heartbeats corrupts session counters under concurrency and issues N+1 sequential queries per batch
## Bug Description

`/api/heartbeats` (the endpoint the VS Code / Cursor coding-presence extension posts to, in batches of up to 25) has two distinct defects in its per-heartbeat loop (`src/app/api/heartbeats/route.ts:109-156`):

**1. Non-atomic counters (read-modify-write).** For each active heartbeat it reads then writes:

```ts
const { data: existing } = await sb
  .from("developer_sessions")
  .select("total_heartbeats, active_seconds")
  .eq("developer_id", dev.id)
  .eq("session_id", hb.sessionId)
  .single();                                   // line 126-131

const { error } = await sb.from("developer_sessions").upsert({
  ...
  total_heartbeats: (existing?.total_heartbeats ?? 0) + 1,
  active_seconds: (existing?.active_seconds ?? 0) + hb.activeSeconds,
}, { onConflict: "developer_id,session_id" });  // line 133-147
```

Two concurrent flushes for the same `session_id` both read the same base value; the second overwrite loses the first's increment. This is the same TOCTOU class already fixed elsewhere in the codebase with atomic RPCs (`grant_xp_atomic`, `increment_visit_count`, `record_mission_progress`) — the heartbeat counters were missed.

**2. N+1 sequential I/O.** The loop `for (const hb of heartbeats)` (line 109) `await`s a SELECT **and** an upsert per item, serially. A full 25-item batch performs up to 50 sequential round-trips to Supabase inside a single serverless invocation, on the app's most frequent write path — multiplying latency and Vercel function duration.

## Steps to Reproduce

**Lost counters:**
1. Open two editor windows (or let the extension retry an in-flight batch) so two requests with the same `session_id` reach `/api/heartbeats` concurrently.
2. Observe `total_heartbeats` / `active_seconds` in `developer_sessions` increments by less than the number of heartbeats sent.

**N+1 latency:**
1. POST a batch of 25 valid heartbeats to `/api/heartbeats`.
2. Observe the handler issues ~50 sequential DB queries and the response time scales linearly with batch size.

## Expected Behavior

- Concurrent heartbeats for the same session should atomically accumulate `total_heartbeats` / `active_seconds` with no lost updates.
- A batch should be persisted without one SELECT + one upsert per item serially.

## Suggested Direction

1. Replace read-then-upsert with a single atomic Postgres RPC, e.g.
   `INSERT ... ON CONFLICT (developer_id, session_id) DO UPDATE SET total_heartbeats = developer_sessions.total_heartbeats + 1, active_seconds = developer_sessions.active_seconds + EXCLUDED.active_seconds, last_heartbeat_at = now()`.
   This removes the SELECT and makes the increment atomic.
2. Eliminate the N+1: aggregate heartbeats per `session_id` in JS first (sum `activeSeconds`, count rows) and issue one RPC per session, or pass the validated array to a single set-returning RPC.
3. Apply the same batching to the offline-update path; keep the single broadcast already done after the loop (line 159).

nyxsky404:
## What does this PR do?

The active branch of `/api/heartbeats` read `total_heartbeats` / `active_seconds` and then wrote `value + 1` per heartbeat. Two flushes for the same session (two editor windows, or a retried batch) both read the same base and the second overwrites the first, so increments get lost. It also ran a SELECT plus an upsert for every heartbeat, so a full 25-item batch was up to 50 sequential round-trips.

Added a `record_heartbeat` RPC (migration `20260621_heartbeat_atomic.sql`) that folds a session's batch into one `INSERT ... ON CONFLICT DO UPDATE` with `total_heartbeats = ... + EXCLUDED.total_heartbeats`. Row-level locking on the update serializes concurrent calls for the same session, so no more lost counts.

In the route, active heartbeats are now aggregated per session (counters summed, descriptive fields last-writer-wins) and persisted with one RPC call per session instead of a query per heartbeat. Offline signals stay a direct update.

## Related issue

Fixes #625

## Screenshots

n/a (backend only)

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/app/api/heartbeats/route.ts
+++ b/src/app/api/heartbeats/route.ts
@@ -106,15 +106,24 @@ export async function POST(request: NextRequest) {
 
   let accepted = 0;
 
-  for (const hb of heartbeats) {
-    // Server generates timestamp, never trust client
-    const now = new Date().toISOString();
-    const isOffline = hb.status === "offline";
+  // Aggregate active heartbeats per session so each session is persisted with
+  // a single atomic RPC (no SELECT-then-upsert per heartbeat). Offline signals
+  // end the session and are applied as direct updates.
+  interface SessionAgg {
+    count: number;
+    activeSeconds: number;
+    language?: string;
+    project?: string;
+    editorName: string;
+    os?: string;
+  }
+  const activeBySession = new Map<string, SessionAgg>();
 
-    if (isOffline) {
+  for (const hb of heartbeats) {
+    if (hb.status === "offline") {
       const { error } = await sb
         .from("developer_sessions")
-        .update({ status: "offline", ended_at: now })
+        .update({ status: "offline", ended_at: new Date().toISOString() })
         .eq("developer_id", dev.id)
         .eq("session_id", hb.sessionId);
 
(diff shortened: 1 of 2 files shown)


Ixotic27 (approved):
This is an exceptional backend performance and concurrency fix. Moving to an aggregated batch design with a single atomic RPC write per session completely eliminates state overwrite race conditions while significantly reducing database round-trips. Outstanding! 🚀

PR #638, opened 2026-06-21
fix(checkin): replace single rate limit with two-layer retry-resilient scheme

### LINKED ISSUE #612: [Bug] Fix Checkin Route Rate Limiting on Transient Failures
**Description:**
The rate limiter on the check-in API route (\/api/checkin\) is currently configured to limit requests to 1 per 5 seconds per user. If a check-in request fails due to transient database issues, network latency, or LeetCode API timeout errors, the user's subsequent immediate retry is blocked with a \429 Too fast\ error.

**Requirements:**
- Implement a retry-resilient rate limiting scheme for \/api/checkin\ (e.g., allow a small burst limit of 3 requests or verify that failed/non-completed check-in attempts do not consume the rate limit window).
- Ensure that rate limits are only enforced for successful or pending operations to prevent blocking user retries on transient errors.

gc-MayankPun:
## What does this PR do?
Replaces the single `rateLimit('checkin:{userId}', 1, 5000)` call with a two-layer approach so transient failures don't block user retries with a 429.

**Before:** one lock fires before any DB work — a failed RPC or LeetCode timeout burns the slot and the retry is blocked for 5s.

**After: two layers with separate keys:**
- **Layer 1 (coarse spam guard):** `checkin:{userId}` — 3 requests per 30s, runs before DB calls. Blocks scripted floods while allowing 2–3 retries on transient failures.
- **Layer 2 (idempotency guard):** `checkin:success:{userId}` — 1 request per 10s, runs only after `perform_checkin` returns `checked_in: true`. Prevents double-submits from double-clicks without ever penalising failed attempts.

Using separate keys ensures a failed request hits layer 1 but never consumes the success bucket, keeping it clean for the retry.

## Related issue
Fixes #612

## Screenshots
N/A — backend-only change.

## Checklist
- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/app/api/checkin/route.ts
+++ b/src/app/api/checkin/route.ts
@@ -104,10 +104,12 @@ export async function POST(request: Request) {
     return NextResponse.json({ error: "Not authenticated" }, { status: 401 });
   }
 
-  // Per-user rate limit: 1 req/5s
-  const { ok } = await rateLimit(`checkin:${user.id}`, 1, 5000);
-  if (!ok) {
-    return NextResponse.json({ error: "Too fast" }, { status: 429 });
+  // Early rate limit: coarse guard against extreme spam (e.g. scripted floods).
+  // Allows 3 requests per 30s so transient failures don't block retries, but
+  // still caps abuse. A tighter idempotency check fires after the RPC succeeds.
+  const { ok: earlyOk } = await rateLimit(`checkin:${user.id}`, 3, 30_000);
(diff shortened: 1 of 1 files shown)


Ixotic27 (approved):
Review: Excellent two-layer rate limit design. It resolves the problem of transient failure lockout while preventing double-checkin spam. Thanks for your contribution!

PR #640, opened 2026-06-21
fix(ticker): replace flex with inline-flex to fix marquee stutter on loop #618

### LINKED ISSUE #618: [Bug]: Infinite scrolling activity ticker stutters and jumps during loop
**Description:**
The Activity Ticker at the bottom of the screen—which acts as an infinite scrolling marquee of recently achieved actions and highlighting stats—stutters, jumps, and glitches instead of looping smoothly.

**Technical Root Cause:**
In [ActivityTicker.tsx](file:///d:/leetcode/src/components/ActivityTicker.tsx), the scrollable element is defined with:
```tsx
<div
  className="ticker-scroll flex whitespace-nowrap"
  style={{ "--ticker-duration": `${Math.max(20, tickerText.length)}s` } as React.CSSProperties}
>
```
The CSS class `flex` compiles to `display: flex`. In block layouts, this causes the container `.ticker-scroll` to stretch to `100%` of its parent's width (the viewport width). 

The CSS animation executes a translation of `-50%`:
```css
@keyframes ticker {
  0% { transform: translateX(0); }
  100% { transform: translateX(-50%); }
}
```
Because the container stretches to `100%` of the viewport rather than matching its content's exact width, a translation of `-50%` moves it by half of the viewport width rather than half of the duplicated content width. This causes the marquee to jump abruptly and stutter instead of looping seamlessly.

**Proposed Solution:**
Modify [ActivityTicker.tsx](file:///d:/leetcode/src/components/ActivityTicker.tsx) to use `inline-flex` (which translates to `display: inline-flex`) so the container wraps its children exactly:
```diff
-        <div
-          className="ticker-scroll flex whitespace-nowrap"
-          style={{ "--ticker-duration": `${Math.max(20, tickerText.length)}s` } as React.CSSProperties}
-        >
+        <div
+          className="ticker-scroll inline-flex whitespace-nowrap"
+          style={{ "--ticker-duration": `${Math.max(20, tickerText.length)}s` } as React.CSSProperties}
+        >
```
Using `inline-flex` sets the container width to `max-content`, ensuring `translateX(-50%)` translates exactly one copy of the duplicated content, resulting in a perfect infinite loop.

---

**Requirements:**
* Update [ActivityTicker.tsx](file:///d:/leetcode/src/components/ActivityTicker.tsx) to use `inline-flex` or equivalent styling so the loop width resolves dynamically based on content width.
* Ensure it works correctly with hover effects (pauses on hover).
* :warning: **Visual Proof Required:** You MUST provide a video screen recording showing the activity ticker running and looping smoothly *before* and *after* applying the fix. PRs without visual proof will be closed.

gc-MayankPun:
## What does this PR do?

Fixes the activity ticker marquee stuttering and jumping during its loop by changing `display: flex` to `display: inline-flex` on the `.ticker-scroll` container.

- With `flex`, the container stretches to 100% of the viewport width, so `translateX(-50%)` translates by half the viewport, not half the content which was causing a visible jump on each loop reset. 
- Switching to `inline-flex` makes the container shrink-wrap to `max-content` width, so `-50%` translates exactly one copy of the duplicated content, producing a seamless infinite loop.

## Related issue

Fixes #618

### ❌ Before — Ticker stutters and jumps on loop

https://github.com/user-attachments/assets/afe5093e-9f49-43ec-aff6-0b413ebd568f

### ✅ After — Ticker loops smoothly with no jump 

https://github.com/user-attachments/assets/0ba58a34-b21c-490e-9179-1767d4ae8c82

## Checklist
- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/components/ActivityTicker.tsx
+++ b/src/components/ActivityTicker.tsx
@@ -98,15 +98,15 @@ export default function ActivityTicker({
 
   return (
     <div
-      className={`fixed ${hasBottomBar ? "bottom-[46px]" : "bottom-0"} sm:bottom-0 left-0 right-0 z-30 flex h-7 items-center border-t border-border/30 bg-bg/90 backdrop-blur-sm`}
+      className={`fixed ${hasBottomBar ? "bottom-11.5" : "bottom-0"} sm:bottom-0 left-0 right-0 z-30 flex h-7 items-center border-t border-border/30 bg-bg/90 backdrop-blur-sm`}
     >
       <div
         className="min-w-0 flex-1 overflow-hidden cursor-pointer"
         onClick={onOpenPanel}
       >
         <div
-          className="ticker-scroll flex whitespace-nowrap"
-          style={{ "--ticker-duration": `${Math.max(20, tickerText.length)}s` } as React.CSSProperties}
+          className="ticker-scroll inline-flex whitespace-nowrap"
+          style={{ "--ticker-duration": `${Math.max(20, tickerText.length * 8)}s` } as React.CSSProperties}
         >
           {[...tickerText, ...tickerText].map((item, i) => (
             <span


Ixotic27 (approved):
Review: This is a clean and correct fix for the activity ticker marquee stutter on loop. Thanks for your contribution!

Ixotic27 (approved):
Review: Clean and correct fix for the activity ticker marquee stutter on loop. Thanks for your contribution!

PR #645, opened 2026-06-21
fix: prune stale arcade_active_players records (stale presence cleanup)

### LINKED ISSUE #603: Multiplayer: Pruning Inactive Active Players (Stale Presence Cleanup)
### Description
When a player disconnects, we call `disconnect()` which deletes their row from `arcade_active_players`. However, if the player suddenly closes the tab, loses network connection, or the browser freezes/sleeps, the active player entry will remain in the database indefinitely.

### Proposed Solution
Implement a pruning mechanism to remove stale active player records.
1. Run a database cron job using pg_cron or an edge function that executes:
```sql
DELETE FROM public.arcade_active_players WHERE last_heartbeat < now() - INTERVAL '1 minute';
```
2. Or perform this cleanup periodically on room fetch/route handler calls to keep the room/player counts accurate.

ionfwsrijan:
## What does this PR do?

Implements stale presence cleanup for `arcade_active_players` — when a player disconnects abruptly (closes tab, loses network, browser freeze/sleeps) their heartbeat row stays in the DB forever. This PR adds three layers of protection:

1. **New Supabase migration** — creates `cleanup_stale_arcade_players()` function scheduled via `pg_cron` every minute to `DELETE` rows where `last_heartbeat < now() - INTERVAL '1 minute'`
2. **New Vercel cron endpoint** (`/api/cron/cleanup-active-players`) — serverless handler with `CRON_SECRET` auth that prunes stale records, registered in `vercel.json` to run every minute
3. **Inline cleanup** in `/api/arcade/rooms/counts` — prunes stale entries before computing live player counts to ensure accuracy regardless of the cron schedule

## Related issue

Fixes #603

## Screenshots

N/A — backend / DB change, no visual diff.

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally


ionfwsrijan:
@Ixotic27 Please review this

Ixotic27:
Hey @ionfwsrijan, thanks for the PR! I checked the changes and noticed two issues:

1. **Database performance / write load on GET request:**
In `src/app/api/arcade/rooms/counts/route.ts`, you have added a delete operation inside a `GET` endpoint:
```typescript
    // Prune stale entries older than 1 minute before counting
    await sb
      .from("arcade_active_players")
      .delete()
      .lt("last_heartbeat", new Date(Date.now() - 60 * 1000).toISOString());
```
Since this endpoint is fetched frequently by all active clients to display live player counts, triggering a database delete query on every single GET request can cause excessive database write load, lock contention, and high latency under load. Since there is already a database cron job (`cleanup_stale_arcade_players`) and a Vercel cron endpoint running every minute to clean up stale records, this prune operation should be removed from the GET endpoint.

2. **Timing-safe comparison crash risk:**
In the cleanup cron route (`src/app/api/cron/cleanup-active-players/route.ts`), the `timingSafeEqual` function compares string lengths before performing a timing-safe buffer comparison:
```typescript
function timingSafeEqual(a: string, b: string): boolean {
  if (a.length !== b.length) return false;
  return crypto.timingSafeEqual(Buffer.from(a), Buffer.from(b));
}
```
If a request is sent with an Authorization header containing multibyte UTF-8 characters of the same string length as the expected header, their byte lengths will differ. Because `crypto.timingSafeEqual` throws a `RangeError` if the input buffers do not have matching byte lengths, this will cause the serverless function to crash.
To fix this, you should check the byte length of the converted buffers instead:
```typescript
function timingSafeEqual(a: string, b: string): boolean {
  const bufA = Buffer.from(a);
  const bufB = Buffer.from(b);
  if (bufA.byteLength !== bufB.byteLength) return false;
  return crypto.timingSafeEqual(bufA, bufB);
}
```

Could you please address these two issues? Thanks!

ionfwsrijan:
@Ixotic27 Please review this now. I've made the changes you asked.

ionfwsrijan:
@Ixotic27 Does it need any changes?

Ixotic27 (approved):
Approved! This is a clean and correct implementation. Thank you for your contribution to LeetCode City!

PR #713, opened 2026-06-25
fix(checkout): use last x-forwarded-for entry for rate-limit IP

### LINKED ISSUE #702: [BUG]: support/checkout uses spoofable first IP for rate limiting
## Bug Description

`src/app/api/support/checkout/route.ts` extracts the client IP using `x-forwarded-for.split(",")[0]` (the first entry), which is the most easily spoofable position. An attacker can send `X-Forwarded-For: 1.2.3.4` to bypass the defense-in-depth rate limit on the checkout route.

The middleware (`src/middleware.ts`) already fixed this exact issue by using the **last** entry (`ips[ips.length - 1]`), with an explicit comment explaining why the first entry is spoofable. The `support/checkout` route still uses the old pattern.

## Steps to Reproduce

1. Send a POST to `/api/support/checkout` with `X-Forwarded-For: 1.2.3.4` header
2. The rate limit key becomes `checkout:1.2.3.4` (spoofed)
3. Repeated requests with different spoofed IPs bypass the rate limit entirely

## Expected Behavior

The route should use the same IP extraction as the middleware: `ips[ips.length - 1]`.

## Acknowledgements

- [x] I have checked the [existing issues](https://github.com/Ixotic27/The-Leetcode-City/issues) and this has not already been reported.
- [x] I have checked the [CONTRIBUTING](https://github.com/Ixotic27/The-Leetcode-City/blob/main/CONTRIBUTING.md) guide and am following the coding standards.

vipul674:
## What does this PR do?
Fixes the support checkout route using a spoofable IP extraction for rate limiting.

The checkout route extracted the client IP using `x-forwarded-for.split(",")[0]` (the first entry), which is the most easily spoofable position. An attacker can send `X-Forwarded-For: 1.2.3.4` to bypass the rate limit.

Now uses the **last** entry (`ips[ips.length - 1]`), matching the existing pattern in `src/middleware.ts` which already has an explicit comment explaining why the first entry is spoofable.

## Related issue
Fixes #702

## Checklist
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] I have tested this change locally
- [x] My code follows the style guidelines of this project

### DIFF
--- a/src/app/api/support/checkout/route.ts
+++ b/src/app/api/support/checkout/route.ts
@@ -6,10 +6,12 @@ import { rateLimit } from "@/lib/rate-limit";
 const MIN_AMOUNT = 10; // minimum ₹10 INR for checkouts
 
 export async function POST(request: Request) {
-  const ip =
-    request.headers.get("x-forwarded-for")?.split(",")[0]?.trim() ??
-    request.headers.get("x-real-ip") ??
-    "unknown";
+  const forwarded = request.headers.get("x-forwarded-for");
+  let ip = request.headers.get("x-real-ip") ?? "unknown";
+  if (forwarded) {
+    const ips = forwarded.split(",").map((s) => s.trim()).filter(Boolean);
+    if (ips.length > 0) ip = ips[ips.length - 1];
+  }
 
   const { ok } = await rateLimit(`checkout:${ip}`, 1, 5_000);
   if (!ok) {


vipul674:
@Ixotic27 Ready for review. IP extraction now uses the last x-forwarded-for entry, matching the existing middleware pattern.

Ixotic27 (approved):
Approved. Safe extraction of the rate-limit IP from x-forwarded-for to handle multiple proxy headers correctly.

PR #724, opened 2026-06-26
fix(shop): insert pending purchase before deducting points and granting item

### LINKED ISSUE #708: [Bug]: buy-with-points grants consumable item before purchase record is written — item not rolled back on INSERT failure
### Bug Description

In `src/app/api/shop/buy-with-points/route.ts`, the order of operations is:

1. **Deduct points atomically** (line ~97) — correct
2. **Grant the item** via `fulfillItemPurchase` (line 111) — writes to DB immediately (increments `streak_freezes_available` or calls `grant_consumable` RPC)
3. **Insert purchase record** (lines 115–126)
4. **If INSERT fails** → rollback *points only* via `add_points_atomic` (lines 132–135), then return `500`

The item grant in step 2 is **irreversible** — `fulfillItemPurchase` calls the `grant_streak_freeze` or `grant_consumable` RPCs directly, and no compensating `revoke_*` function exists anywhere in the codebase. When the purchase INSERT fails (transient Postgres error, connection pool exhaustion, or a `idempotency_key` conflict on retry), the error handler restores the user's points but **cannot un-grant the item**.

Result: the user ends up with their original points **plus** the consumable they never paid for.

**Relevant code:**

`src/app/api/shop/buy-with-points/route.ts` lines 111–137:
```ts
// Step 2 — item granted here, BEFORE purchase record exists
const { status: purchaseStatus } = await fulfillItemPurchase(dev.id, item_id, admin);

// Step 3 — purchase record written after fulfillment
const { data: purchase, error: purchaseError } = await admin
  .from("purchases")
  .insert({ ..., status: purchaseStatus })
  .select("id")
  .single();

// Step 4 — rollback only restores points; item already granted
if (purchaseError) {
  if (!isDev) {
    await admin.rpc("add_points_atomic", {
      p_developer_id: dev.id,
      p_price_points: item.price_points,
    });
  }
  return NextResponse.json({ error: "Failed to record purchase" }, { status: 500 });
}
```

`src/lib/items.ts` lines 245–288 — `fulfillItemPurchase` for consumables (no rollback path exists):
```ts
if (itemId === "streak_freeze") {
  const { error: freezeError } = await sb.rpc("grant_streak_freeze", { p_developer_id: developerId });
  // grant_streak_freeze increments streak_freezes_available — not reversible
}
// BATTLE_CONSUMABLES follow same pattern via grant_consumable RPC
```

A secondary path: the `idempotency_key` is `points_${dev.id}_${item_id}_${Date.now()}`. Using `Date.now()` means the key is **not stable across retries**. A network timeout after the purchase INSERT succeeded but before the response arrived causes a retry with a new key — the deduction + fulfillment execute again, creating a duplicate purchase record and a second item grant, against an already-depleted balance.

**Note:** This is distinct from the concurrent-race issue previously reported in #690 (two concurrent requests). This is a sequential single-request failure path — no concurrency required.

### Steps to Reproduce

1. Buy a consumable item (e.g., `streak_freeze`) via the shop with enough points
2. Simulate or wait for a transient Supabase INSERT failure at the purchase record step (can be reproduced in staging by temporarily introducing a unique constraint violation on `idempotency_key` via an intercepting proxy, or by observing Supabase connection errors during a deploy)
3. Observe the API returns `500 "Failed to record purchase"`
4. Check: user's points are restored AND `streak_freezes_available` is still incremented

### Expected Behavior

Either:
- **Option A:** Write the purchase record first (with `status: 'pending'`), then fulfill the item, then update to `status: 'completed'`. A failure during fulfillment leaves the record in `pending` and no item is granted.
- **Option B:** Move the entire operation into a single Postgres stored procedure that atomically deducts points, records the purchase, and grants the item.
- **Option C:** Add `revoke_streak_freeze` and `revoke_consumable` RPCs and call them in the `purchaseError` catch block alongside `add_points_atomic`.

Additionally, the `idempotency_key` should be a deterministic hash of `(developer_id, item_id, request_fingerprint)` rather than including `Date.now()`.

N/A — the issue is observable by reading `src/app/api/shop/buy-with-points/route.ts` lines 111–137 and `src/lib/items.ts` lines 219–289 together. The absence of any `revoke_consumable` or `revoke_streak_freeze` function can be confirmed by searching the codebase.

### Acknowledgements

nyxsky404:
## What does this PR do?

Fixes the pre-grant atomicity gap in `/api/shop/buy-with-points`:

**Before:** `fulfillItemPurchase` ran first (step 4), before the purchase record was written (step 5) and before points were deducted (step 6). A failed INSERT left the user with the item AND their original points.

**After:** The operation is now ordered correctly:
1. INSERT purchase record with `status: 'pending'`
2. Deduct points atomically (delete pending record on failure — no item touched yet)
3. `fulfillItemPurchase` (restore points + delete pending record on failure)
4. UPDATE purchase to final status

### Idempotency key

The idempotency key is now **supplied by the client** instead of generated on the server. A server-generated value (whether `Date.now()` or `crypto.randomUUID()`) is recreated on every request, so a network-timeout retry produces a brand-new key and bypasses duplicate detection entirely.

- The client generates the key once per purchase attempt and sends it in the `Idempotency-Key` header.
- The server validates the header, namespaces it per developer + item (`points_<dev>_<item>_<clientKey>`), and relies on the existing unique index on `purchases.idempotency_key`.
- A unique-violation (`23505`) on insert is treated as an idempotent retry and returns `200` with the current balance, instead of a `500`.

This keeps the key stable across retries of the *same* operation, so a timeout-and-retry can no longer create a duplicate purchase record or double-grant an item.

## Related issue

Fixes #708

## Screenshots

N/A — logic-only change.

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally


Ixotic27:
Hi @nyxsky404,

Thanks for working on this! I noticed a logical issue with how the `idempotencyKey` is generated in `src/app/api/shop/buy-with-points/route.ts`:

- The key is being generated on the server side using `crypto.randomUUID()` inside the POST request handler.
- If a client experiences a network timeout or connection drop and retries the request, a new POST request will hit the server and generate a brand-new UUID.
- This completely defeats the purpose of the idempotency key, which is intended to detect retries of the *same* operation. The key must originate from the client (e.g., generated on the client side and passed in the headers or body).

Please update it to receive the idempotency key from the client request.

Ixotic27:
Hi @nyxsky404, thanks for the fix!
Generating \crypto.randomUUID()\ on the server side inside the route handler defeats retry protection.
If a client request times out and the client retries, the server will receive a new request and run the handler again, generating a *new* random UUID. Since the idempotency key changes, it will bypass duplicate checks.
For idempotency to work correctly, the unique transaction key/UUID should be generated by the client and sent in the request headers (or request body), and the server should check/validate that client-provided key.

Ixotic27:
Hi @nyxsky404, there is an idempotency design issue here: you generate the idempotencyKey on the server using crypto.randomUUID(). Because a client retry starts a new request, a new UUID is generated every time. This defeats the purpose of the idempotency key (preventing duplicate purchase records on retry). To be effective, the key must be supplied by the client or generated from client-provided properties that remain stable across retries.

nyxsky404:
Thanks for the review @Ixotic27 — you're right, a server-generated key defeats retry protection. Fixed:

- The idempotency key is now **generated on the client** and sent via the `Idempotency-Key` header (`ShopClient.tsx`).
- The server validates the header, namespaces it per developer + item, and treats a unique-key violation (`23505`) as an idempotent retry returning `200` instead of `500`.

So a timeout-and-retry now reuses the same key and can no longer create a duplicate purchase or double-grant an item. Also fixed the `lint-build` failure (escaped some entities + matched the file-level eslint-disable convention already used in `page.tsx`). PR description updated to reflect the new approach.

Ixotic27 (approved):
Looks solid! Merging now. Thanks for your contribution! 🚀

PR #731, opened 2026-06-27
[Bug]: Fix serializeDeveloper omitting numeric fields with value 0

### LINKED ISSUE #722: [Bug]: serializeDeveloper omits numeric fields with value 0 causing undefined in consumer code
### Describe the bug

`serializeDeveloper` in `src/lib/serialize.ts` strips fields where the value equals `0` for ~20+ numeric fields (current_week_contributions, rating, streak counts, LeetCode metrics, GitHub stats, etc.). Consumers that receive the serialized snapshot get `undefined` instead of `0` for these fields.

This causes silent `NaN` bugs when client code performs arithmetic on those fields (e.g., `dev.rating + 100` → `NaN`). It also breaks strict equality checks (`dev.current_streak === 0` → `false` when field is missing).

### To Reproduce

1. Load a city snapshot that runs `serializeDeveloper` on developers
2. Access a numeric field that happens to be `0` (e.g., `current_week_contributions`, `easy_solved`, `rating`)
3. The field returns `undefined` instead of `0`
4. Any arithmetic or comparison against `0` behaves incorrectly

### Expected behavior

`serializeDeveloper` should **not** strip `0` values from numeric fields. Removing `0` values for bandwidth optimization is premature — it shifts the correctness burden to every consumer. Either:

1. Remove all `if (val === 0) continue;` lines for numeric fields, OR
2. Keep the optimization but provide a client-side helper that fills default `0` values for missing numeric fields

### Additional context

- File: `src/lib/serialize.ts`
- Affected fields: `kudos_count`, `visit_count`, `app_streak`, `raid_xp`, `current_week_contributions`, `current_week_kudos_given`, `current_week_kudos_received`, `xp_total`, `easy_solved`, `medium_solved`, `hard_solved`, `acceptance_rate`, `contest_rating`, `lc_streak`, `followers`, `following`, `organizations_count`, `current_streak`, `longest_streak`, `active_days_last_year`, `language_diversity`, `total_prs`, `total_reviews`, `total_issues`, `repos_contributed_to` (and more)
- Same issue applies to boolean fields stripped when `false` (`claimed`, `rabbit_completed`, `district_chosen`)
- Used by city-snapshot API route at `src/app/api/cron/city-snapshot/route.ts`

vipul674:
## What does this PR do?

Removes all `if (val === 0) continue;` and `if (val === false) continue;` branches from `serializeDeveloper` in `src/lib/serialize.ts`. These branches stripped ~20+ numeric fields (rating, streaks, LeetCode metrics, GitHub stats) when their value was `0`, causing consumers to receive `undefined` instead of `0` — leading to `NaN` bugs in arithmetic and broken equality checks.

Also removes the `xp_level === 1` and `building_style === "tower"` default-value stripping for consistency.

## Related issue

Fixes #722

## Screenshots

N/A

## Checklist

- [ ] `npm run lint` passes
- [ ] Tested locally
- [ ] No secrets or .env values committed
- [ ] I acknowledge that an automated AI Reviewer will perform a preliminary review of this PR.
- [ ] ⭐ **I have starred this repository!** (We prioritize PRs and assignments for stargazers)


vipul674:
@Ixotic27 PR ready for review.

**Summary:** Removes all default-value stripping branches from `serializeDeveloper` that dropped fields when their value was `0` (numeric) or `false` (boolean). Previously ~20+ fields (rating, streaks, LeetCode metrics, GitHub stats) would return `undefined` instead of `0`, causing `NaN` bugs in consumer arithmetic.

**Changes:** Removed 31 lines of `if (key === "..." && val === 0/false/1/"tower") continue;` branches.

vipul674:
**Fix for blocked CI:** Replaced `any` with `unknown` in `serializeDeveloper` signature to satisfy `@typescript-eslint/no-explicit-any` lint rule.

vipul674:
@Ixotic27 CI checks all green now. Ready for review.

Ixotic27:
Hi @vipul674, thanks for the contribution!

There is a type safety issue in this implementation:
\\\	ypescript
const val = dev[key] as Record<string, unknown>;
\\\
Casting \al\ as \Record<string, unknown>\ globally is a type assertion bypass because \al\ can be a primitive (number, string, boolean). A cleaner approach is to keep \al\ typed properly and perform the type assertion or check only inside the \	ypeof val === 'object'\ block where object properties like \loadout\ or \ctive_raid_tag\ are accessed.

vipul674:
@Ixotic27 Fixed the type safety issue. Changed `const val = dev[key] as Record<string, unknown>` to keep `val` as `unknown` and only cast inside the `typeof val === "object"` block via `const obj = val as Record<string, unknown>`. Branch also updated to latest main.

Ixotic27:
Hello @vipul674! Thank you for the contribution to fix the serialization defaults issue.

We noticed a critical runtime bug in your changes:
In `src/lib/serialize.ts`, you removed the `&& val !== null` check when matching objects:
```ts
if (typeof val === "object") {
  const obj = val as Record<string, unknown>;
  if (key === "loadout") {
    const isDefault = !obj.crown ...
```
In JavaScript/TypeScript, `typeof null` returns `"object"`. If `loadout` or `active_raid_tag` is `null` in the database, this block will execute and throw a TypeError: `Cannot read properties of null (reading 'crown')`, crashing the API endpoint.

Please restore the null check:
```ts
if (typeof val === "object" && val !== null) {
```
Let us know when you update the PR!

Ixotic27:
@vipul674 thanks for the PR! There are two critical issues with these changes:

1. **Runtime Crash (Uncaught TypeError)**: JavaScript's `typeof null` evaluates to `"object"`. By removing the `&& val !== null` check, if a developer has a null database field (such as `active_raid_tag`), it will pass the check and evaluate `!obj.attacker_login`, which throws a runtime exception and crashes the entire serialization route.
2. **Performance/Payload Size**: Removing the default-value omission rules defeats the optimization of `serializeDeveloper`. It increases the API payload size by 3x-4x for thousands of developers, adding significant load time and memory overhead.

Ixotic27:
@vipul674 Friendly reminder: please restore the `&& val !== null` check to prevent the runtime crash on null fields, and make sure default serialization omission optimization is preserved. Thank you!

Ixotic27 (approved):
Great job fixing serializeDeveloper so it doesn't omit valid zero values!

PR #878, opened 2026-07-03
fix: merge XP grant into redeem_xp_code RPC for atomicity

### LINKED ISSUE #863: [Bug] XP redeem code consumed before XP is granted — non-atomic operation can lose XP
### Bug Description

The XP redeem code flow in `src/app/api/shop/redeem-xp/route.ts` has a non-atomic two-step process: first the code is consumed (via `redeem_xp_code` RPC), then XP is granted (via `grant_xp_atomic` RPC). If the second RPC fails, the code is permanently consumed but the developer receives no XP.

**Sequence in `src/app/api/shop/redeem-xp/route.ts`:**
1. Lines 71-76: `redeem_xp_code` RPC — atomically records usage (`xp_code_usages` insert) and increments `used_count`. If this succeeds, the code is consumed and cannot be reused.
2. Lines 108-112: `grant_xp_atomic` RPC — applies XP to the developer's account. This is a **separate** network call in a **separate** database transaction.

If the server crashes, the database connection drops, or `grant_xp_atomic` returns an error between steps 1 and 2, the code is permanently consumed but the XP is never granted. The error handler at lines 114-119 correctly logs the failure and returns a "Contact support" message, but the damage is done — the code slot is gone and manual DB intervention is required to recover.

By contrast, the item redeem code flow (likely in a separate route) should be used as a reference for atomicity: the code usage and reward grant should happen in the same database transaction.

### Steps to Reproduce

1. Obtain a valid XP redeem code with remaining uses
2. Send a redeem request
3. Intercept/timing: let `redeem_xp_code` RPC succeed (code is consumed)
4. Force `grant_xp_atomic` to fail (e.g., DB connection timeout, constraint violation)
5. The response is: `"Code was redeemed but XP could not be applied. Contact support."`
6. The code cannot be reused (it was consumed in step 3), and the XP was never credited

### Expected Behavior

The code consumption and XP grant should be atomic — either both succeed or both fail. The `redeem_xp_code` RPC should also apply the XP within the same database function/transaction, eliminating the window between the two operations. Alternatively, the route could use a two-phase approach where:
- The code usage is recorded in a "pending" state
- XP is granted within the same DB transaction
- The code usage is finalized

This ensures that a crash or failure between steps cannot orphan a consumed-but-unrewarded code.

### Acknowledgements

- [X] I have searched existing issues to make sure this is not a duplicate.
- [X] I have read the CONTRIBUTING.md guidelines.
- [X] ⭐ I have starred this repository! (We prioritize issue assignments for stargazers)

vipul674:
## What does this PR do?

Fixes #863 — XP redeem code flow had non-atomic two-step process: code consumption then XP grant in separate transactions. Failure between steps orphaned consumed codes.

## Related issue

Fixes #863

## Changes

- **`supabase/migrations/061_xp_redeem_code_atomic_grant.sql`**: New migration — `redeem_xp_code` RPC now calls `grant_xp_atomic` internally within same transaction. Rolls back on grant failure.
- **`src/app/api/shop/redeem-xp/route.ts`**: Removed separate `grant_xp_atomic` call, reads `new_total`/`new_level` from RPC result directly.

## Checklist

- [x] I have read the CONTRIBUTING.md guidelines
- [x] `npm run lint` passes
- [x] I have starred this repository


vipul674:
@Ixotic27 PR #878 ready for review. Fixes XP redeem non-atomicity by merging the `grant_xp_atomic` call into the `redeem_xp_code` RPC so code consumption and XP grant happen in a single transaction. Includes new migration `061_xp_redeem_code_atomic_grant.sql` and updated route. If XP grant fails, the entire transaction rolls back.

Ixotic27:
Hi @vipul674! Thank you for working on this. 

We noticed a minor issue: in the POST handler you destructured the RPC response as `rpcResult`, and removed the helper variable `const result = rpcResult?.[0];`. However, the downstream code still references the undefined `result` variable (e.g. `result?.error_code`, `result.xp_amount`, `result.new_total`, and `result.new_level`). 

This will throw a `ReferenceError` at runtime. Please restore `const result = rpcResult?.[0];` or update the references to use `rpcResult?.[0]` to fix this. Thank you!

vipul674:
@Ixotic27 the `const result = rpcResult?.[0];` is actually already present at current HEAD (line 66 of route.ts). The third commit (merge from main) re-added it after the second commit's refactor. No `ReferenceError` on the current state. Thanks for the thorough review 🙌

Ixotic27 (approved):
LGTM! The atomic redeem and XP grant changes look solid. Merging.

PR #879, opened 2026-07-03
fix: add row-count verification to Rabbit Quest sightings 1-4

### LINKED ISSUE #862: [Bug] Rabbit Quest allows concurrent sighting progression — race condition in sightings 1-4 bypasses sequential order guard
### Bug Description

The Rabbit Quest POST handler in `src/app/api/rabbit/route.ts` has a race condition for sightings 1-4. While each request verifies `sighting === (current_progress + 1)`, concurrent requests that both read the same progress value can both pass this check, allowing a user to skip sightings or complete the quest faster than intended.

The final sighting (5) has proper optimistic locking via `.eq("rabbit_completed", false)` with row-count verification, but sightings 1-4 only use `.lt("rabbit_progress", sighting)` without checking the number of affected rows. This means concurrent requests for different sighting values can both report success even though only one UPDATE actually committed.

**Root cause in `src/app/api/rabbit/route.ts`:**
- Line 48: Sequential check `sighting !== (dev.rabbit_progress ?? 0) + 1` is evaluated client-side from stale data
- Lines 120-126: `.lt("rabbit_progress", sighting)` guards against regression but doesn't prevent concurrent forward advancement
- Lines 128-132: No row-count check on the UPDATE result — 0 affected rows still returns `{ progress: sighting, completed: false }`

### Steps to Reproduce

1. Have a developer with `rabbit_progress = 2`
2. Send concurrent POST requests with `sighting: 3` and `sighting: 4` simultaneously
3. Both requests read `rabbit_progress = 2`
4. Request A (sighting=3): passes check 3===2+1 → UPDATE with `.lt("rabbit_progress", 3)` succeeds, sets progress to 3
5. Wait, actually request B would fail since 4 !== 3. A more realistic scenario: send `sighting: 3` and `sighting: 3` simultaneously after `rabbit_progress = 2`

Concurrent sighting-3 scenario:
1. Two requests both read `rabbit_progress = 2`
2. Both pass `3 === 2 + 1`
3. Request A UPDATE `.lt("rabbit_progress", 3)` → succeeds, progress=3
4. Request B UPDATE `.lt("rabbit_progress", 3)` → evaluates `3 < 3` = false → 0 rows affected
5. Request B still returns `{ progress: 3, completed: false }` — no error, no indication that the update was silently skipped

While progress 3 is idempotent, the lack of row-count verification means a client cannot distinguish between "my update was applied" and "someone else's update won". Combined with the rate limiter (2 req/s), this enables a subtle race where rapid concurrent requests can lead to inconsistent client state.

### Expected Behavior

All 5 sightings should use the same pattern: optimistic lock with row-count verification. For sightings 1-4, the handler should:
1. Use `.eq("rabbit_progress", sighting - 1)` instead of `.lt("rabbit_progress", sighting)` as the WHERE guard
2. Check `data.length` (or use a RETURNING clause) to confirm the UPDATE actually changed a row
3. Return a proper error response when 0 rows are affected (another concurrent request advanced first)

This ensures exactly one request per sighting value commits, matching the pattern already used for sighting 5.

### Acknowledgements

- [X] I have searched existing issues to make sure this is not a duplicate.
- [X] I have read the CONTRIBUTING.md guidelines.
- [X] ⭐ I have starred this repository! (We prioritize issue assignments for stargazers)

vipul674:
## What does this PR do?

Fixes #862 — Rabbit Quest sightings 1-4 lacked row-count verification, allowing concurrent requests to silently skip progression.

## Related issue

Fixes #862

## Changes

- **`src/app/api/rabbit/route.ts`**: Changed `.lt("rabbit_progress", sighting)` to `.eq("rabbit_progress", sighting - 1)` with `.select("id")` row-count check. Returns 409 when concurrent request already advanced. Also incorporated upstream main's extension auth support in GET handler.

## Checklist

- [x] I have read the CONTRIBUTING.md guidelines
- [x] `npm run lint` passes
- [x] I have starred this repository

### DIFF
--- a/src/app/api/rabbit/route.ts
+++ b/src/app/api/rabbit/route.ts
@@ -4,10 +4,6 @@ import { getSupabaseAdmin } from "@/lib/supabase";
 import { rateLimit } from "@/lib/rate-limit";
 import { getAuthenticatedDeveloper } from "@/lib/arena";
 
-// POST - Record rabbit sighting encounter
-/**
- * @param {import('next/server').NextRequest} request
- */
 export async function POST(request: Request) {
   const supabase = await createServerSupabase();
   const {
@@ -44,16 +40,11 @@ export async function POST(request: Request) {
     return NextResponse.json({ progress: 5, completed: true });
   }
 
-  // Must be sequential: no skipping
   if (sighting !== (dev.rabbit_progress ?? 0) + 1) {
     return NextResponse.json({ error: "Wrong sighting order" }, { status: 400 });
   }
 
   if (sighting === 5) {
-    // Final sighting: complete the quest.
-    // Use an optimistic-lock WHERE clause so that only the first of any
-    // concurrent sighting-5 requests commits; the rest see rowCount = 0
-    // and return early before touching purchases.
     const { data, error } = await admin
       .from("developers")
       .update({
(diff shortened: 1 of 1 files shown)


Ixotic27 (approved):
LGTM! The code changes look clean and correct. Merging.

PR #1037, opened 2026-07-21
refactor(city): extract shared city projection service (#1024)

### LINKED ISSUE #1024: [Feature]: City projection logic is duplicated across API and snapshot pipelines
### Feature Description

The current city data pipeline derives and serializes the same developer payload through multiple code paths. The API response builder, snapshot generation, and serialization utilities each participate in shaping the city payload, resulting in overlapping transformation and serialization responsibilities.

This feature proposes introducing a shared server-side projection layer that becomes the single source of truth for city payload construction while preserving the existing API contract and runtime behavior.

### Problem it Solves

The city payload is currently assembled through multiple related paths, including API responses, snapshot generation, and serialization utilities. Because these responsibilities are spread across separate modules, future changes to the city data model or serialization rules may require updates in several places.

This increases maintenance effort and makes it easier for the different data paths to diverge over time. Consolidating the projection logic would reduce duplication, improve consistency, and make future schema evolution significantly easier.

### Alternatives Considered

An alternative approach would be to keep the existing structure and continue updating each projection path independently whenever the city payload changes.

While this avoids a refactor, it continues to duplicate mapping and serialization responsibilities across multiple modules, increasing the long-term maintenance cost.

Another option would be to share only serialization helpers while leaving projection logic separate. However, this would only partially address the duplication and would still leave multiple sources of truth for constructing the city payload.

### Additional Context

This proposal is intended as an architectural improvement rather than a behavioral change.

The goal is to preserve the existing API response shape while consolidating duplicated projection responsibilities into a shared implementation that can be reused by the API, snapshot generation, and other server-side consumers.

The expected benefits include:

* A single source of truth for city payload construction
* Reduced duplication across the city data pipeline
* Easier maintenance when the city schema evolves
* Lower risk of inconsistencies between API responses and generated snapshots
* Improved testability of the projection layer

### Acknowledgements

ZainabTravadi:
### What does this PR do?

This PR refactors the city data pipeline by introducing a shared server-side projection layer for developer payload construction.

### Changes

* Added a shared `cityProjection` module to centralize developer payload shaping.
* Refactored the city serializer to delegate projection logic to the shared implementation.
* Updated the city snapshot pipeline to reuse the same projection layer.
* Removed duplicated projection and serialization logic while preserving the existing API response and snapshot formats.
* Added focused regression tests for the shared projection layer.

This keeps the city projection logic as a single source of truth, reducing maintenance overhead and minimizing the risk of drift between API and snapshot generation.

---

## Related issue

Fixes #1024

---

## Screenshots

N/A — Backend architectural refactor (no UI changes).

---

## Checklist

* [x] `npm run lint` passes
* [x] Tested locally
* [x] No secrets or `.env` values committed
* [x] I acknowledge that an automated AI Reviewer will perform a preliminary review of this PR.
* [x] ⭐ **I have starred this repository!** (We prioritize PRs and assignments for stargazers)


Ixotic27 (approved):
Reviewed and approved. Clean refactoring of developer projection logic into a dedicated, reusable cityProjection service.

PR #1247, opened 2026-07-31
✨ feat: add Cache-Control no-store to arcade favorites API route

### LINKED ISSUE #1190: feat : add Cache-Control header to arcade favorites API route

The `GET /api/arcade/favorites` route does not set a `Cache-Control` header. For a public read-only endpoint, adding appropriate cache headers reduces unnecessary server load and improves client performance.

- In `src/app/api/arcade/favorites/route.ts`, add a `Cache-Control: no-store` header to the `NextResponse` returned from the `GET` function
- This ensures the favorites list is always fresh and not cached

Prevents stale data issues for users' favorites list and reduces unnecessary database queries from cached responses. Consistent with cache header patterns used in other routes like `leaderboard-position` and `stats`.

Anijesh:
## What does this PR do?

Adds `Cache-Control: no-store` to `/api/arcade/favorites` so a user's favorites list is never stored by the browser or a shared CDN.

While implementing this I found the route only exported a `POST` handler (the favorite toggle) — the `GET` endpoint named in the issue did not exist yet. The favorites list is currently served as a side-channel of `/api/arcade/rooms` and read in `src/app/arcade/page.tsx` as `data.favorites`. So this PR delivers both halves of the issue's intent:

**1. `Cache-Control: no-store` on every response path**
Added via a shared `NO_STORE` constant so no return path is missed — success, `400 room_id required`, `401 Unauthorized`, `500`, and the catch-block fallback all send the header.

**2. Added the `GET /api/arcade/favorites` handler the issue describes**
Auth-gated with `resolveAuthenticatedDeveloper({ loadDeveloper: false })` (the same helper `POST` already uses), returning the signed-in user's `room_id` list from `arcade_room_favorites`. A missing/unreachable arcade table degrades to `{ favorites: [] }`, matching the existing fallback in `/api/arcade/rooms` rather than breaking the room browser.

The route stays `ƒ` (dynamic, server-rendered on demand) in the build output, so no `export const dynamic` declaration is needed — it reads cookies via `createServerSupabase`.

`no-store` is the right directive here rather than the `s-maxage` used by `leaderboard-position` and `stats`: those are public aggregates, whereas favorites are per-user and change the instant a room is toggled. This matches the existing `no-store` routes: `loadout`, `presence`, `search`, and `arcade/rooms/counts`.

### Tests

Added `src/app/api/arcade/favorites/route.test.ts` — 8 tests following the mocking style of the neighbouring `arcade/avatar/route.test.ts`. It covers `GET` (list, unauthenticated, query error, table-unreachable fallback) and `POST` (add, remove, validation error, unauthenticated), asserting `Cache-Control: no-store` on each path.

```
Test Files  1 passed (1)
     Tests  8 passed (8)
```

### Note for maintainers (not changed here, out of scope)

`/api/arcade/rooms` returns the per-user `favorites` and `recentVisits` arrays under `Cache-Control: public, s-maxage=10, stale-while-revalidate=30`. Because those fields are user-specific, a shared cache can serve one user's favorites to another. Happy to open a separate issue/PR for that if you'd like — I left it alone to keep this PR scoped to #1190.

## Related issue

Fixes #1190

## Screenshots

Not applicable — backend-only change, no visual surface.

## Verification

```
npx vitest run src/app/api/arcade/favorites/route.test.ts   # 8/8 passed
npx eslint src/app/api/arcade/favorites/                    # clean, 0 problems
npx tsc --noEmit                                            # clean
npm run build                                               # passes (CI env vars)
```

The 9 failing tests elsewhere in the suite (`api/city`, `webhooks/stripe`, `lib/notification-helpers`) are pre-existing on `main` — verified they fail identically with these changes stashed.

## Checklist

- [x] `npm run lint` passes — clean on the changed files (`npx eslint src/app/api/arcade/favorites/` reports 0 problems). The repo-wide run still reports its existing errors in unrelated files, unchanged by this PR.
- [x] Tested locally


Anijesh:
Thanks for the review — all three inline comments plus the suppressed one were valid, and I've addressed each in `99a1886`. I verified every one against the code before changing anything, and added a regression test for each behavioural fix.

### 1. `req.json()` could throw past the `no-store` guarantee ✅ fixed

Correct, and it undercut the whole point of this PR. `req.json()` sat outside the `try`, so a malformed body escaped the handler as a bodyless 500 with no `Cache-Control` header at all — the one thing the PR claims can't happen. Now parsed inside a `try` that returns a `400 Invalid JSON body` with `NO_STORE`.

Confirmed against the unfixed route, where the exception escapes rather than producing a response:

```
× does not allow a malformed JSON body to escape uncached
  SyntaxError: Expected property name or '}' in JSON at position 2
```

### 2. Delete path ignored the Supabase error ✅ fixed

Also correct. A failed delete still returned `{ favorited: false }`, desyncing the client from a row that was still in the table. It now surfaces the error as a `500` with `NO_STORE`, matching what the insert path already did.

```
× does not report success when the delete fails
  AssertionError: expected 200 to be 500
```

This one was pre-existing rather than introduced here, but it's two lines and sits in the path this PR touches, so it seemed better to fix than to leave behind.

### 3. Delete mock wasn't thenable ✅ fixed

Good catch — `await` on a plain object is a no-op, so the test would have passed even if the route stopped awaiting the delete. The builder is now chainable via `.eq()` *and* thenable, which is also what lets the new delete-failure test above work at all.

### 4. (suppressed) Test didn't assert the delete actually ran ✅ fixed

Added assertions that the delete was called and scoped to both `user_id` and `room_id`, so the route can't regress to reporting `favorited: false` without deleting anything.

### Verification

```
npx vitest run src/app/api/arcade/favorites/route.test.ts   # 10/10 passed (was 8)
npx eslint src/app/api/arcade/favorites/                    # 0 problems
npx tsc --noEmit                                            # clean
npm run build                                               # exit 0, ✓ Compiled successfully
```

Both new tests fail against the unfixed route, as shown above.

Anijesh:
Hi @Ixotic27 — no rush at all, I can see `main` has needed a lot of attention the last couple of days.

Whenever the queue opens up, this one is ready whenever you are:

- Production Build and lint-build both pass on the current head
- Labels are in order and the GSSoC label audit is green
- All four Copilot review points are addressed in `99a1886`, each with a regression test that fails against the pre-fix code

Two of those were real bugs rather than style nits — `req.json()` could throw straight past the `no-store` guarantee this PR is about, and a failed delete was being reported to the client as success. Both are covered by tests now.

Thanks for merging `main` in along the way — happy to rebase, split this up, or drop the added `GET` handler into its own PR if any of that would make it easier to review.

Ixotic27 (approved):

PR #1248, opened 2026-07-31
🐛 fix: improve error handling in lc-pulse POST route

### LINKED ISSUE #1189: fix : improve error handling in lc-pulse POST route

The `POST /api/lc-pulse` route uses `err: any` type annotation in the catch block and accesses `err.message` directly without checking if the error is an `Error` instance. This is a TypeScript type safety issue.

- In `src/app/api/lc-pulse/route.ts`, change the catch block parameter from `err: any` to a properly typed `err: unknown`
- Use `err instanceof Error ? err.message : "Internal server error"` for the error response
- Add a `Cache-Control: no-store` header to the response

Fixes TypeScript type safety issues, improves robustness when non-Error exceptions are thrown, and ensures the response is not cached by CDNs. Consistent with error handling patterns used in other routes in the codebase.

Anijesh:
## What does this PR do?

Fixes the unsafe top-level `catch` in `POST /api/lc-pulse` and stops the route's responses from being cached.

While implementing this I confirmed the `catch (err: any)` is **a real runtime bug, not only a typing nit** — and the new tests demonstrate it. Running them against the unmodified file produces:

```
TypeError: Cannot read properties of null (reading 'message')
```

Because `err.message` was read without narrowing:

- a thrown **non-`Error`** (e.g. a string) made `err.message` `undefined`, so the client received a bare `{}` with a 500;
- a thrown **`null`/`undefined`** made `err.message` raise a `TypeError` *inside the catch block*, which escaped the handler entirely and produced Next.js's bodyless 500 — no JSON at all.

### Changes

**1. Typed the catch parameter as `unknown` and narrowed it**
`err instanceof Error ? err.message : "Internal server error"`, as specified in the issue, so every thrown value produces a well-formed JSON response.

**2. Added a log line consistent with the rest of the file**
The two inner catches already log via `console.warn("[app/api/lc-pulse/route.ts] ...", err)`; the top-level catch was silently swallowing the error with no trace. Now matches that existing prefix.

**3. Added `Cache-Control: no-store`**
Applied via a shared `NO_STORE` constant across all four response paths — `401`, `404`, `500`, and success — since this is an authenticated, per-user presence endpoint. The route remains `ƒ` (dynamic, server-rendered on demand) in the build output.

### Tests

Added `src/app/api/lc-pulse/route.test.ts` — 6 tests following the mocking style of the neighbouring `arcade/avatar/route.test.ts`, covering success, `401`, `404`, a thrown `Error`, a thrown non-`Error`, and a thrown `null`, asserting `Cache-Control: no-store` on every path.

Two of them are direct regression tests for the bugs above. **All 6 fail against the pre-change file** — verified by stashing the fix and re-running:

```
× reports presence and forbids caching on success
× does not allow unauthenticated responses to be cached
× does not allow the unlinked-account response to be cached
× surfaces the message when a real Error is thrown
× falls back to a generic message when a non-Error value is thrown
× still returns JSON when a nullish value is thrown
    TypeError: Cannot read properties of null (reading 'message')
```

With the fix applied:

```
Test Files  1 passed (1)
     Tests  6 passed (6)
```

### Note for maintainers (not changed here, out of scope)

`getRecentSubmissions` still has `submissions.filter((s: any) => ...)`. The file originally carried two `@typescript-eslint/no-explicit-any` errors — this PR removes the one the issue names (`err: any`) and leaves the other, since it sits in a different function and is unrelated to error handling. Happy to fold that one-line fix in if you'd prefer it here rather than as a separate issue.

## Related issue

Fixes #1189

## Screenshots

Not applicable — backend-only change, no visual surface.

## Verification

```
npx vitest run src/app/api/lc-pulse/route.test.ts   # 6/6 passed
npx eslint src/app/api/lc-pulse/                    # no new errors (see note above)
npx tsc --noEmit                                    # clean
npm run build                                       # passes; /api/lc-pulse listed as ƒ (dynamic)
```

The 9 failing tests elsewhere in the suite (`api/city`, `webhooks/stripe`, `lib/notification-helpers`) are pre-existing on `main` — verified they fail identically with these changes stashed.

## Checklist

- [x] `npm run lint` passes — no errors introduced on any line this PR touches (CI lints changed lines only). The file's remaining pre-existing `any` is described in the note above.
- [x] Tested locally


Anijesh:
Hi @Ixotic27 — no rush, just flagging that this one is green whenever the queue opens up.

- Production Build and lint-build both pass on the current head
- Copilot reviewed it and marked it **🟢 Ready to approve** with zero comments

Worth noting the `catch (err: any)` this fixes was a genuine runtime bug, not only a typing issue: a thrown `null` made `err.message` raise a `TypeError` *inside* the catch block, which escaped the handler and produced a bodyless 500. There's a regression test for that case, plus one for non-`Error` throws.

Happy to rebase or adjust anything if it helps. Thanks!

Ixotic27 (approved):

PR #1446, opened 2026-08-05
🐛 fix(ci): resolve race condition between auto-labeler and gssoc validator

### LINKED ISSUE #1249: [Bug]: GSSoC Label Audit intermittently races the Auto-Labeler and fails on newly-opened PRs
## Bug Description

**GSSoC Label Audit** fails on every newly-opened PR, reporting labels as missing that the Auto-Labeler adds seconds later. The check then has no way to correct itself, so it stays red and emails the contributor a failure — even though the PR is correctly labelled.

`GSSoC Label Validator` and `PR Auto Labeler & Review` both trigger on the same `pull_request_target` event, so they start **in parallel**. The validator finishes in ~3s and reads an unlabelled PR; the labeler applies the labels ~10-20s later.

The validator *does* listen for `labeled` events, which should fix this — but [`gssoc-label-validator.yml` L21-23](https://github.com/Ixotic27/The-Leetcode-City/blob/main/.github/workflows/gssoc-label-validator.yml#L21-L23) skips runs whose actor is `github-actions[bot]`:

```yaml
# Skip bot-triggered label events unless it's a review submission to avoid loops
if: >
  github.actor != 'github-actions[bot]' ||
  github.event_name == 'pull_request_review'
```

The Auto-Labeler calls `addLabels` with `secrets.GITHUB_TOKEN` ([`pr-auto-labeler.yml` L203](https://github.com/Ixotic27/The-Leetcode-City/blob/main/.github/workflows/pr-auto-labeler.yml#L203)), so the `labeled` event it produces has actor `github-actions[bot]` and is skipped. The stale failure can never be re-evaluated on its own.

## Evidence

Timings for PR #1247, from the Actions API:

| Time (UTC) | Event |
|---|---|
| `18:41:23` | `GSSoC Label Validator` and `PR Auto Labeler` both start |
| `18:41:26` | Validator concludes **failure** after 3s — 0 labels visible |
| `18:41:41` | Auto-Labeler applies all 6 labels — 15s too late |

The same pattern on PR #1248: both started `19:05:01`, validator failed, labels applied `19:05:09`.

Across the last 100 `pull_request_target` runs there were 16 `GSSoC Label Validator` runs with 3 failures — every one of them the *first* run on a fresh PR:

```
2026-07-31T19:05:01Z  4b1f11a  1189-lc-pulse-error-handling
2026-07-31T18:41:23Z  782582e  1190-cache-control-arcade-favorites
2026-07-31T17:59:25Z  12c3716  #1093-change
```

The third is another contributor's branch, so this is not specific to one account. Every validator run that came *after* a push to an already-labelled PR succeeded.

## Steps to Reproduce

1. Open a PR from a fork against `main` that links an approved GSSoC issue.
2. Watch the **GSSoC Label Audit** check — it fails within ~3s with "Missing `gssoc:approved`", "Missing difficulty label", "Missing type label".
3. Check the PR's labels a few seconds later — all three are present.
4. The check stays red. Adding or removing a label as the bot does not re-trigger it.

## Expected Behavior

The label audit should evaluate the PR's labels *after* the Auto-Labeler has applied them, and a stale failure should clear itself once the labels are present — without the contributor having to force-push.

## Current Workaround

Push a new commit (or amend + force-push). That produces a `synchronize` event whose actor is the contributor rather than the bot, so the validator runs and passes. Confirmed on both #1247 (`782582e` → `16fb5a2`) and #1248 (`4b1f11a` → `f6e366a`).

This works but is noisy: it costs every contributor a red X, a failure email, and a pointless force-push.

## Suggested Fix

Either of these should resolve it:

**Option A — let bot-driven label events through.** The validator only calls `createComment` / `updateComment` / `setFailed`; it never adds or removes a label, so it cannot trigger a `labeled` event and cannot loop through this path. Narrowing the guard to still process label events would be safe:

```yaml
if: >
  github.event_name == 'pull_request_review' ||
  github.event.action == 'labeled' ||
  github.event.action == 'unlabeled' ||
  github.actor != 'github-actions[bot]'
```

**Option B — order the two workflows.** Have the audit run on `workflow_run` after `PR Auto Labeler & Review` completes, so it never reads a pre-label snapshot.

Option A is the smaller change and also fixes the case where a maintainer's manual re-label should re-validate.

I'm happy to open a PR for whichever approach you prefer — I've already traced the workflows while debugging this on my own PRs.

## Environment

- Affects: `.github/workflows/gssoc-label-validator.yml`, `.github/workflows/pr-auto-labeler.yml`
- Not browser-specific — CI/workflow behaviour only.

atharv96k:
## What does this PR do?

Updates the `jobs.validate-labels.if` guard in `.github/workflows/gssoc-label-validator.yml` to allow bot-driven `labeled` and `unlabeled` events to execute.

- Resolves the race condition where `gssoc-label-validator` evaluated pre-labeled snapshots on PR creation and failed permanently.
- Allows `github-actions[bot]` label events to re-audit the PR without causing workflow loops.

## Related issue

Fixes #1249

## Checklist

- [x] `npm run lint` / formatting passes
- [x] Tested locally

### DIFF
--- a/.github/workflows/gssoc-label-validator.yml
+++ b/.github/workflows/gssoc-label-validator.yml
@@ -20,8 +20,10 @@ jobs:
     runs-on: ubuntu-latest
     # Skip bot-triggered label events unless it's a review submission to avoid loops
     if: >
-      github.actor != 'github-actions[bot]' ||
-      github.event_name == 'pull_request_review'
+      github.event_name == 'pull_request_review' ||
+      github.event.action == 'labeled' ||
+      github.event.action == 'unlabeled' ||
+      github.actor != 'github-actions[bot]'
 
     steps:
       - name: Audit GSSoC Labels


Ixotic27 (approved):

PR #1449, opened 2026-08-05
🐛 fix: return 500 instead of fake success when arcade favorite toggle throws

### LINKED ISSUE #1245: [Bug]: Arcade favorites API reports success when the database write fails
### Bug Description

`POST /api/arcade/favorites` can return a 200 success response when the database write did not actually happen, causing a favorite to silently revert on the next page load.

There are two distinct paths in `src/app/api/arcade/favorites/route.ts`:

**1. The remove-favorite path discards its error (lines 31-35).**

The delete is awaited, but its result is never destructured or checked, and the route unconditionally returns `{ favorited: false }`:

```ts
await sb.from("arcade_room_favorites")
  .delete()
  .eq("user_id", user.id)
  .eq("room_id", room_id);
return NextResponse.json({ favorited: false });
```

Five lines below, the add-favorite path *does* check its error and returns a 500. So the two halves of the same toggle handle failure inconsistently:

```ts
const { error } = await sb.from("arcade_room_favorites")
  .insert({ user_id: user.id, room_id });

if (error) {
  return NextResponse.json({ error: error.message }, { status: 500 });
}
```

**2. The catch block returns a hardcoded `favorited: true` (lines 47-50).**

The existing comment names this behaviour directly:

```ts
} catch (e) {
  console.warn("Could not toggle favorite in DB, fallback to fake success:", e);
  // Just toggle locally
  return NextResponse.json({ favorited: true });
}
```

Because the value is hardcoded, an *un*-favorite that throws responds `{ favorited: true }` — the opposite of the action the user took, and the opposite of what the optimistic UI is displaying.

Both paths matter because the client trusts the status code. `src/app/arcade/page.tsx:133` rolls back its optimistic update only via `if (!res.ok) setFavorites(prev)`. A 200 means the rollback never fires, so the star stays lit while nothing was persisted.

This is not a duplicate of #673, which fixed a *client-side* stale-snapshot race on rapid toggles. This report is about server-side error handling.

### Steps to Reproduce

1. Sign in and go to `/arcade`.
2. Click the star on any room to favorite it. This succeeds and persists correctly.
3. Simulate a failing delete: in `src/app/api/arcade/favorites/route.ts`, temporarily change line 34 from `.eq("room_id", room_id)` to `.eq("room_id_typo", room_id)` so the delete returns a PostgREST error.
4. Click the same star again to un-favorite it.
5. Open the Network tab and inspect the request — it returns **200** with `{ "favorited": false }`, and the star clears in the UI.
6. Refresh the page.
7. The room is **still favorited**. The delete never happened, but the API reported success and the UI showed it as removed.

### Expected Behavior

The route should only report success when the database write actually succeeded.

- The remove-favorite path should destructure and check its error, matching the add path directly below it — `const { error } = await sb...delete()...` — and return a 500 when `error` is set.
- The catch block should return a 500 rather than a fabricated success, so the client's `if (!res.ok)` rollback fires and the star returns to its true state.
- The response should never report `favorited: true` for an un-favorite action.

With that in place, step 7 above would show the star still lit (matching the database) instead of silently diverging from it.

_No response_

### Browser

_No response_

### Acknowledgements

Anijesh:
## What does this PR do?

Stops `POST /api/arcade/favorites` from reporting success when the database write never happened.

### Scope note — half of this issue is already fixed on `main`

The issue describes two paths. Before writing anything I checked both against current `main`, and **path 1 is already resolved**: #1247 (merged earlier today) added the `deleteError` check to the remove-favorite path, along with the `NO_STORE` headers. Re-fixing it would be a no-op.

So this PR addresses **only the second path — the catch block**, which is still live on `main`:

```ts
} catch (e) {
  console.warn("Could not toggle favorite in DB, fallback to fake success:", e);
  // Just toggle locally
  return NextResponse.json({ favorited: true }, { headers: NO_STORE });
}
```

### Why the hardcoded value is wrong twice over

1. It claims a write that did not happen. `src/app/arcade/page.tsx:133` rolls back its optimistic update only via `if (!res.ok) setFavorites(prev)`, so a 200 means the rollback never fires and the star stays lit against a row that was never written.
2. Because the value is **hardcoded**, an *un*-favorite that throws responds `{ favorited: true }` — the opposite of the action the user took.

### The fix

The catch now returns a 500 so the client's existing rollback fires:

```ts
console.error("Could not toggle favorite in DB:", e);
return NextResponse.json(
  { error: "Could not toggle favorite" },
  { status: 500, headers: NO_STORE }
);
```

I kept the `NO_STORE` header that #1247 established, and left the `GET` fallback alone on purpose — degrading a *read* to "no favorites" keeps the room browser alive, but a *write* cannot be faked. That asymmetry is now stated in the comment so the fallback doesn't get re-added later.

`console.warn` → `console.error`, since this is now a real error path rather than a swallowed one.

### Tests

Two cases added to the existing `route.test.ts`, one per failure mode above. The `createToggleAdmin` helper gained `throws` / `deleteThrows` options so the mock can reject the way an unreachable table actually does.

**I verified the new tests fail without the fix** rather than assuming they were meaningful — stashing only `route.ts` and re-running:

```
× does not fabricate success when the favorites table is unreachable
  AssertionError: expected 200 to be 500
× never reports favorited: true for an un-favorite that throws
  AssertionError: expected 200 to be 500
Tests  2 failed | 10 passed (12)
```

Both pass with the fix — 12/12.

The second test asserts on the response **body**, not just the status, because the status alone would not have caught the inverted-action half of the bug.

## Related issue

Fixes #1245

## Screenshots

Not applicable — API error-handling change with no visual surface.

## Verification

```
npx vitest run src/app/api/arcade/favorites/route.test.ts   # 12/12 passed
npx eslint src/app/api/arcade/favorites/                    # clean, 0 problems
npm run build                                               # ✓ Compiled successfully (exit 0)
```

`src/app/arcade/page.tsx:128` is the only caller of this endpoint, and it already handles `!res.ok`, so no client change is needed.

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/app/api/arcade/favorites/route.test.ts
+++ b/src/app/api/arcade/favorites/route.test.ts
@@ -48,9 +48,13 @@ function createListAdmin({ data = [], error = null, throws }: SelectResult) {
 function createToggleAdmin({
   existing,
   deleteError = null,
+  throws,
+  deleteThrows,
 }: {
   existing: FavoriteRow | null;
   deleteError?: { message: string } | null;
+  throws?: Error;
+  deleteThrows?: Error;
 }) {
   const insert = vi.fn().mockResolvedValue({ error: null });
   const del = vi.fn();
@@ -64,8 +68,13 @@ function createToggleAdmin({
       delEq(...args);
       return deleteBuilder;
     },
-    then: (resolve: (value: { error: { message: string } | null }) => unknown) =>
-      Promise.resolve({ error: deleteError }).then(resolve),
+    then: (
+      resolve: (value: { error: { message: string } | null }) => unknown,
+      reject?: (reason: unknown) => unknown
+    ) =>
+      deleteThrows
+        ? Promise.reject(deleteThrows).then(resolve, reject)
+        : Promise.resolve({ error: deleteError }).then(resolve, reject),
   };
 
   mockFrom.mockImplementation((table: string) => {
@@ -76,7 +85,9 @@ function createToggleAdmin({
     const query = {
       select: () => query,
       eq: () => query,
-      single: vi.fn().mockResolvedValue({ data: existing, error: null }),
+      single: vi.fn().mockImplementation(() =>
+        throws ? Promise.reject(throws) : Promise.resolve({ data: existing, error: null })
+      ),
       delete: () => {
         del();
         return deleteBuilder;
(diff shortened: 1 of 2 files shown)


Ixotic27 (approved):

Issue #190, opened 2026-05-31
[Bug]: Raid history API accepts invalid pagination limits


DhruvalBhinsara1:
### Bug Description
The raid history endpoint accepts `limit` values directly with `Math.min(50, parseInt(...))`. Invalid or negative values such as `?limit=abc` and `?limit=-5` are not normalized to a safe lower/default bound.

### Steps to Reproduce
1. Open `src/app/api/raid/history/route.ts`.
2. Check the `limit` parsing line: `Math.min(50, parseInt(searchParams.get("limit") ?? "20", 10))`.
3. Request `/api/raid/history?developer_id=1&limit=abc` or `/api/raid/history?developer_id=1&limit=-5`.
4. The computed `limit` becomes `NaN` or a negative number and is passed into Supabase `.range(offset, offset + limit - 1)` and JavaScript `.slice(0, limit)`.

### Expected Behavior
The API should use a safe pagination limit, matching the feed endpoint behavior: default invalid values to 20, clamp the minimum to 1, and clamp the maximum to 50.

### Screenshots / Screen Recordings
Not applicable. This is a backend pagination validation bug visible from code inspection.

### Browser
Other

### Acknowledgements
- [x] I have searched existing issues to make sure this is not a duplicate.
- [x] I have read the CONTRIBUTING.md guidelines.
- [x] ⭐ I have starred this repository! (We prioritize issue assignments for stargazers)

### Proposed Solution
Add a small pagination parser for raid history, clamp `limit` to `1..50`, default invalid limits to `20`, keep `offset` non-negative, and add focused tests so negative/NaN values cannot reach the database range calculation.

DhruvalBhinsara1:
/assign gssoc

### Proposed Solution
I will normalize raid history pagination by defaulting invalid limits to 20, clamping limits to 1–50, keeping offsets non-negative, and adding focused tests for negative/NaN query values before opening the linked PR.

Srejoye:
assign me

### Proposed Solution

**Technical Assessment**

Line 10 of `src/app/api/raid/history/route.ts`:

```ts
const limit = Math.min(50, parseInt(searchParams.get("limit") ?? "20", 10));
```

`parseInt("abc", 10)` returns `NaN`. `Math.min(50, NaN)` returns `NaN`. So `limit` becomes `NaN`, which then propagates into two places:

[...]

PR #1450, opened 2026-08-05
🐛 fix: guard NaN and negative pagination in four API routes

### LINKED ISSUE #1246: [Bug]: Four API routes still accept NaN and negative pagination values
### Bug Description

`Math.max(1, NaN)` evaluates to `NaN` in JavaScript, so clamping with `Math.min` / `Math.max` without an explicit `isNaN` guard silently passes `NaN` straight through into the database query.

This exact defect has already been reported and fixed twice — #486 (arena leaderboard) and #190 (raid history). That is why `src/lib/parse-pagination.ts` exists, and why `src/app/api/arena/leaderboard/route.ts:16-19` now guards correctly:

```ts
const rawLimit = parseInt(searchParams.get("limit") ?? "", 10);
const limit = Number.isNaN(rawLimit) ? 100 : Math.min(100, Math.max(1, rawLimit));
```

However, four routes were never migrated and still use the unguarded pattern:

| Route | Line | Problem |
|---|---|---|
| `src/app/api/arcade/rooms/route.ts` | 10-11 | `page` and `limit` both become `NaN` |
| `src/app/api/arcade/maps/route.ts` | 20-21 | `page` and `limit` both become `NaN` |
| `src/app/api/arcade/leaderboard/route.ts` | 19 | `NaN`, **and** no lower bound — accepts negatives |
| `src/app/api/pixels/history/route.ts` | 29-32 | `limit` becomes `NaN` |

Verified in Node:

```js
Math.max(1, parseInt("abc", 10))                // NaN
Math.min(50, Math.max(1, parseInt("abc", 10)))  // NaN
Math.min(parseInt("-5", 10), 50)                // -5   <-- no lower clamp
```

In `arcade/rooms` the `NaN` then propagates into `offset = (page - 1) * limit`, producing `.range(NaN, NaN)` on the Supabase query rather than the intended defaults.

The correct guarded logic already exists in-repo as `parsePagination` in `src/lib/parse-pagination.ts`, and is unit-tested via `src/app/api/raid/history/route.test.ts` — but it is currently imported by only one route.

### Steps to Reproduce

1. Start the app with `npm run dev` (port 3001).
2. Request a room list with a non-numeric page and limit:
   ```
   curl "http://localhost:3001/api/arcade/rooms?page=abc&limit=abc"
   ```
3. `page` and `limit` both evaluate to `NaN`, so the route issues `.range(NaN, NaN)` instead of falling back to the intended defaults of page 1 / limit 20.
4. Request a negative limit on the arcade leaderboard:
   ```
   curl "http://localhost:3001/api/arcade/leaderboard?limit=-5"
   ```
5. `-5` passes through unclamped, because line 19 applies only `Math.min(..., 50)` with no lower bound.
6. For contrast, send the same inputs to `/api/arena/leaderboard?limit=abc` — that route correctly falls back to its default, because it was fixed in #486.

### Expected Behavior

Malformed pagination input should fall back to sane defaults instead of reaching the database, consistent with how #486 and #190 were resolved.

- `?page=abc` → defaults to page 1
- `?limit=abc` → defaults to the route's documented default (20, or 10 for arcade leaderboard)
- `?limit=-5` → clamps to the minimum of 1
- `?limit=99999` → clamps to the existing maximum of 50

The cleanest fix is to route all four through the existing `parsePagination` helper rather than repeating the guard inline in each file. Its `[1, 50]` limit clamp already matches three of the four routes; `arcade/leaderboard` uses a default of 10, so it would pass its own `defaultLimit` argument.

_No response_

### Browser

_No response_

### Acknowledgements

Anijesh:
## What does this PR do?

Routes all four unguarded endpoints through the existing `parsePagination` helper, so malformed pagination input falls back to sane defaults instead of reaching the database as `NaN`.

I confirmed all four were still unguarded on current `main` before starting.

| Route | Before | After |
|---|---|---|
| `api/arcade/rooms` | `page`/`limit` → `NaN`, then `.range(NaN, NaN)` | `parsePageParams` |
| `api/arcade/maps` | `page`/`limit` → `NaN`, then `.range(NaN, NaN)` | `parsePageParams` |
| `api/arcade/leaderboard` | `NaN`, **and** no lower bound | `parsePagination(raw, null, 10)` |
| `api/pixels/history` | `limit` → `NaN` | `parsePagination(raw, null)` |

### Why a second helper was needed

The issue suggests routing all four through `parsePagination`, and two of them fit it directly — `arcade/leaderboard` and `pixels/history` are `limit`-only (the latter is cursor-paginated), so they pass `null` for offset and read `.limit`.

`arcade/rooms` and `arcade/maps` don't fit as cleanly: they're **page**-based, not offset-based, and both echo `page` back in the response body. `parsePagination` neither parses a page nor returns one. Inlining a page guard in both files would reintroduce exactly the duplication this issue is about, so I added a small page-based counterpart in the same module:

```ts
export function parsePageParams(
  rawPage: string | null,
  rawLimit: string | null,
  defaultLimit = 20
): { page: number; limit: number; offset: number } {
  const parsedPage = parseInt(rawPage ?? "", 10);
  const page = Number.isNaN(parsedPage) ? 1 : Math.max(1, parsedPage);

  const { limit } = parsePagination(rawLimit, null, defaultLimit);

  return { page, limit, offset: (page - 1) * limit };
}
```

It **delegates the limit to `parsePagination`** rather than re-deriving it, so the `[1, 50]` clamp and the NaN fallback stay defined in exactly one place — the whole point of #486 and #190. `offset` is only computed after both inputs are known non-NaN, which is what stops `.range(NaN, NaN)`.

### Behaviour matches the issue's expected behaviour exactly

- `?page=abc` → page 1
- `?limit=abc` → 20, or 10 for `arcade/leaderboard` (passed as `defaultLimit`)
- `?limit=-5` → 1 — this is the new lower bound for `arcade/leaderboard`, which previously applied only `Math.min(..., 50)`
- `?limit=99999` → 50

One extra case the inline pattern also got wrong: `?page=` (present but empty) returns `""`, not `null`, so the old `?? "1"` default never applied and it became `NaN` too. That now resolves to 1.

`pixels/history` needed no `defaultLimit` argument — the helper's default of 20 and `[1, 50]` clamp already match what that route applied inline, so its behaviour is unchanged for valid input.

### Tests

15 cases added to `src/lib/__tests__/parse-pagination.test.ts` covering `parsePageParams`: the page guard (NaN / null / empty / `-5` / `0`), limit delegation, derived offset, and the regression itself — that `parsePageParams("abc", "abc")` yields `{page: 1, limit: 20, offset: 0}` rather than `NaN`.

No existing test changed. The suite still asserts the original `parsePagination` contract, which this PR does not alter.

## Related issue

Fixes #1246

## Screenshots

Not applicable — backend input-validation change with no visual surface.

## Verification

```
npx vitest run src/lib/__tests__/parse-pagination.test.ts   # 41/41 passed
npx vitest run src/app/api/arcade/ src/app/api/pixels/ \
    src/lib/__tests__/parse-pagination.test.ts \
    src/app/api/raid/history/route.test.ts                  # 68/68 passed, 6 files
npx eslint <all 6 changed files>                            # clean, 0 problems
npm run build                                               # ✓ Compiled successfully (exit 0)
```

`api/raid/history` is included above because it is the existing `parsePagination` consumer — its behaviour is unchanged.

## Checklist

- [x] `npm run lint` passes
- [x] Tested locally

### DIFF
--- a/src/app/api/arcade/leaderboard/route.ts
+++ b/src/app/api/arcade/leaderboard/route.ts
@@ -1,5 +1,6 @@
 import { NextRequest, NextResponse } from "next/server";
 import { getSupabaseAdmin } from "@/lib/supabase";
+import { parsePagination } from "@/lib/parse-pagination";
 
 type ScoreRow = {
   user_id: string;
@@ -16,7 +17,9 @@ type DevLoginRow = {
 // GET /api/arcade/leaderboard?game=10s_classic&limit=10
 export async function GET(req: NextRequest) {
   const game = req.nextUrl.searchParams.get("game") ?? "10s_classic";
-  const limit = Math.min(parseInt(req.nextUrl.searchParams.get("limit") ?? "10", 10), 50);
+  // This route is offset-less, so only `limit` is used. Its default is 10
+  // rather than the helper's 20, hence the explicit third argument.
+  const { limit } = parsePagination(req.nextUrl.searchParams.get("limit"), null, 10);
 
   const sb = getSupabaseAdmin();
   let data: ScoreRow[] = [];
--- a/src/app/api/arcade/maps/route.ts
+++ b/src/app/api/arcade/maps/route.ts
@@ -1,6 +1,7 @@
 import { NextRequest, NextResponse } from "next/server";
 import { getSupabaseAdmin } from "@/lib/supabase";
 import { resolveAuthenticatedDeveloper } from "@/lib/authenticated-developer";
+import { parsePageParams } from "@/lib/parse-pagination";
 
 function slugify(text: string): string {
   return text
@@ -17,9 +18,10 @@ export async function GET(req: NextRequest) {
   const search = url.searchParams.get("q")?.trim();
   const category = url.searchParams.get("category");
   const creatorId = url.searchParams.get("creator_id");
-  const page = Math.max(1, parseInt(url.searchParams.get("page") ?? "1", 10));
-  const limit = Math.min(50, Math.max(1, parseInt(url.searchParams.get("limit") ?? "20", 10)));
-  const offset = (page - 1) * limit;
+  const { page, limit, offset } = parsePageParams(
+    url.searchParams.get("page"),
+    url.searchParams.get("limit")
+  );
 
   const sb = getSupabaseAdmin();
 
--- a/src/app/api/arcade/rooms/route.ts
+++ b/src/app/api/arcade/rooms/route.ts
@@ -1,5 +1,6 @@
 import { NextRequest, NextResponse } from "next/server";
 import { getSupabaseAdmin } from "@/lib/supabase";
+import { parsePageParams } from "@/lib/parse-pagination";
 
 // GET /api/arcade/rooms — list rooms with search, category, pagination
 export async function GET(req: NextRequest) {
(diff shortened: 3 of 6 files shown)


Ixotic27 (approved):

Issue #1627, opened 2026-08-07
[Bug]: fix: remove remaining city emojis from script console log messages


kaur-jass:
## Summary

There are still several console log messages in the `scripts/` directory that include the city emoji (`🏙️`).

To keep the repository consistent with the project's no-emoji policy, these log messages should use plain text instead.

## Files

- scripts/backfill-lc-profiles.ts
- scripts/lc-hourly-fetcher.ts
- scripts/seed-lc-infinite.ts
- scripts/seed-lc-mass.ts
- scripts/seed-lc.ts

## Proposed Changes

- Remove the `🏙️` emoji from the console log messages.
- Keep the existing wording and functionality unchanged.
- Only update the displayed text.

## Impact

- Consistent with the project's no-emoji policy.
- Cleaner console output.
- No functional changes.

### Steps to Reproduce

1. Clone the repository and open it in VS Code.
2. Navigate to the `scripts/` directory.
3. Open the following files:
   - `scripts/backfill-lc-profiles.ts`
   - `scripts/lc-hourly-fetcher.ts`
   - `scripts/seed-lc-infinite.ts`
   - `scripts/seed-lc-mass.ts`
   - `scripts/seed-lc.ts`
4. Observe that the startup `console.log` messages contain the `🏙️` emoji.

### Expected Behavior

The remaining console log messages in the `scripts/` directory should not contain emojis.

All startup and status logs should use plain text (or a consistent text prefix such as `[LC City]`) to align with the project's no-emoji policy while preserving the existing functionality.

### Acknowledgements

- [x] I have searched existing issues to make sure this is not a duplicate.
- [x] I have read the CONTRIBUTING.md guidelines.
- [x] ⭐ I have starred this repository! (We prioritize issue assignments for stargazers)

kaur-jass:
### Proposed Solution

Hi, @Ixotic27 

I would like to work on this issue.

My approach:

- Identify all remaining console log messages in the `scripts/` directory that still include the city emoji (`🏙️`).
- Replace the emoji with plain-text messages (or a consistent text prefix such as `LC City` or `[LC City]`) while preserving the existing wording.
- Ensure the changes are limited to log messages only, with no modifications to the script logic or functionality.
- Verify that all scripts continue to run normally and display the updated console output.

This change will make the repository more consistent with the project's no-emoji policy and improve the uniformity of console messages without affecting behavior.

Kindly assign this issue to me.

/assign

[...]

PR #1718, opened 2026-08-17
fix: remove city emoji from script console log messages

aditya226-sharma:
Removed 🏙️ emoji from console.log messages in scripts/. Relates to #1627


aditya226-sharma:
@Ixotic27 Gentle ping — ready for review!

